PR #19, opened 2026-05-07
ci: GitHub Actions Tests workflow — pytest matrix + mypy + gitleaks (closes #13)

timon0305:
First in-repo CI workflow for cppa-cursor-browser.

## What this gates

- **unittest** — 3 OSes × 3 Pythons = 9 cells (`[ubuntu-latest, macos-latest, windows-latest]` × `[3.11, 3.12, 3.13]`). Required.
- **typecheck (mypy)** — Python 3.12 / ubuntu-latest. Codebase already has ~70 typed functions, so mypy does real work. Lenient flags (`--ignore-missing-imports --no-strict-optional`) + step-level `continue-on-error: true` until the surface is clean.
- **secret-scan (gitleaks)** — full git history walk, version 8.21.2 with SHA-256 verification. Mirrors the gitleaks setup used on the-claw.

## Why this scope

1. Python version drift is real (3.11 → 3.12 dict ordering / 3.12 → 3.13 PEP 695 / etc.) — the matrix catches it cheap
2. Project has type hints across most of the surface; mypy can run today, no project adoption step needed
3. Secret-scan is language-agnostic and zero-friction

What this PR deliberately does NOT add:
- **lint (ruff / black)** — needs the project to adopt a config first (`ruff.toml` or `pyproject.toml`); separate decision
- **multi-OS for typecheck/secret-scan** — `mypy` and `gitleaks` are OS-agnostic, multi-OS would be wasted CI minutes

## Test plan

- [x] `python -m unittest discover tests` — **137 tests pass** locally
- [x] `python yaml.safe_load` + `actionlint` clean on the workflow
- [x] All action SHAs pinned (no `@v4`/`@v5` mutable tags)
- [ ] First CI run on this PR will exercise all 9 unittest cells + typecheck + secret-scan; mypy may surface findings in lenient mode (non-blocking via `continue-on-error`)

Closes #13.

## Summary by CodeRabbit

* **Chores**
  * Enhanced CI/CD pipeline with improved security configurations and automated secret detection.
  * Expanded test coverage across Python 3.11–3.13.
  * Added type checking validation to the continuous integration workflow.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -5,11 +5,23 @@ on:
     branches: [master]
   pull_request:
 
-jobs:
-  unittest:
-    name: Unit tests
-    runs-on: ubuntu-latest
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
 
+jobs:
+  # ── Unit tests: matrix across OS and Python version ───────────────────────
+  # Closes #13. The unittest suite is the merge gate. Multi-OS catches the
+  # rare path / line-ending issue that a single-OS run hides; multi-Python
+  # catches API drift across LTS / current / latest interpreters.
+  unittest:
+    name: Unit tests (${{ matrix.os }} / Python ${{ matrix.python-version }})
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        python-version: ["3.11", "3.12", "3.13"]
     steps:
       # Pinned to immutable commit SHAs (not @v4 / @v5) so a compromised tag
       # cannot silently swap the underlying action code on this CI runner.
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -0,0 +1,133 @@
...
+  # rare path / line-ending issue that a single-OS run hides; multi-Python
+  # catches API drift across LTS / current / latest interpreters.
+  unittest:
+    name: Unit tests (${{ matrix.os }} / Python ${{ matrix.python-version }})
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        python-version: ["3.11", "3.12", "3.13"]
```
bradjin8, line 30: Scope / cost: Nine unittest cells plus two extra jobs increase CI time and minutes versus issue #13’s single 3.12 job. Fine if the team accepts that tradeoff; otherwise align with the issue (one OS / one Python) or document the budget change.

Confirm with team

.github/workflows/tests.yml
```diff
@@ -0,0 +1,133 @@
...
+          python-version: "3.12"
+
+      - name: Install runtime deps + mypy
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install 'flask>=3.0' 'fpdf2>=2.7' 'mypy>=1.10'
+
+      - name: Run mypy
+        continue-on-error: true
+        run: mypy --ignore-missing-imports --no-strict-optional --pretty .
```
bradjin8, line 81: continue-on-error: true on mypy: Type errors never fail the workflow, so merges can stay “green” while mypy is broken. Acceptable as a transitional step only; plan to drop continue-on-error (or tighten mypy) once the codebase is clean.

bradjin8 (approved):
Check the comments, both are optional.
Please confirm with the team for cost/scope before moving forward.

.github/workflows/tests.yml
wpak-ai: Since the operator will be interacting with a server deployment (likely Ubuntu), let's remove everything else.
Multiple python versions seem good.

.github/workflows/tests.yml
wpak-ai: Sounds reasonable as a transitional step.

PR #20, opened 2026-05-07
fix: disable debug=True / Werkzeug debugger exposure (closes #9)

timon0305:
## Problem

`app.py` was hard-coding `debug=True` in its `app.run(...)` call. The Werkzeug debugger that ships with Flask debug mode is a documented remote-code-execution primitive — anyone reaching the listening port (and guessing the PIN, which is generated from machine-stable inputs) can execute arbitrary Python in the server process. The default `127.0.0.1` bind is the only thing standing between the operator and RCE; a misconfigured `--host`, an SSH tunnel, or a careless reverse proxy is enough to expose it.

There was also no way to opt out — the literal was unconditional.

## Change

Three files:

- **[`app.py`](app.py)** — removed `debug=True`. Default debug now **off**. Opt-in via either `--debug` CLI flag or `FLASK_DEBUG=1` env var. When debug is enabled, prints a stderr WARNING naming the RCE risk + reminding the operator to bind only to loopback. Auto-reloader gated on the same flag.
- **[`utils/debug_flag.py`](utils/debug_flag.py)** — new tiny module with `resolve_debug_flag(env_value, cli_flag)`. Lives outside `app.py` so it can be unit-tested without importing Flask (matching the existing `_build_app_parser` mirror convention in the test suite). Truthy env values: `"1"`, `"true"`, `"yes"` — case-insensitive, whitespace-tolerant. CLI flag wins.
- **[`tests/test_cli_args.py`](tests/test_cli_args.py)** — new `TestDebugFlagGating` class: 8 cases covering env-truthy / env-falsey / env-unset / CLI-override + argparse `--debug` default & explicit + a **source-level guard** that fails if `debug=True` is ever re-introduced as a literal in `app.py`.

## Test plan

Unit: `python3 -m unittest tests.test_cli_args` → **42/42 OK** (was 34, +8 new).

Live smoke (all four matrix cells, on a real `flask>=3.0` install):

| Trial | Setting | `Debug mode` | WARNING | `Debugger is active!` |
|-------|---------|------|------|------|
| A | (default — no env, no flag) | off | — | — |
| B | `--debug` | on | ✅ | ✅ |
| C | `FLASK_DEBUG=1` | on | ✅ | ✅ |
| D | `FLASK_DEBUG=0` | off | — | — |

Bogus paths in trials A/D return a plain Flask 404 — **not** the orange Werkzeug debugger console. The WARNING prints once per process (parent + reloader child = 2× when the reloader is on, which is correct).

## Severity

Critical — Werkzeug debugger exposure is a documented RCE pathway. Listed as a Critical / 1pt item in the eval week-1 plan for `cppa-cursor-browser`.

Closes #9.

## Summary by CodeRabbit

* **New Features**
  * Debug mode is now off by default; can be enabled explicitly via a new --debug CLI option or FLASK_DEBUG.
  * When enabled, a clear security warning about the remote debugger is shown.

* **Tests**
  * Expanded test coverage for debug flag behavior and added a regression guard preventing always-on debug.

* **Documentation**
  * README updated to document debugger behavior and how to enable it.

### DIFF
--- a/app.py
+++ b/app.py
@@ -1,5 +1,5 @@
 """
-Cursor Chat Browser — Python Edition
+Cursor Chat Browser - Python Edition
 A Flask web application for browsing and managing chat histories
 from the Cursor editor's AI chat feature.
 """
@@ -124,7 +124,7 @@ if __name__ == "__main__":
     if debug_enabled:
         # Print the warning to stderr so it's visible even when stdout is
         # piped/redirected. The Werkzeug debugger is a remote-code-execution
-        # primitive — anyone reaching the host:port can hijack the process.
+        # primitive - anyone reaching the host:port can hijack the process.
         print(
             "WARNING: Flask debug mode ENABLED. The Werkzeug debugger allows "
             "arbitrary code execution by anyone who can reach this server. "
--- a/app.py
+++ b/app.py
@@ -4,11 +4,14 @@ A Flask web application for browsing and managing chat histories
 from the Cursor editor's AI chat feature.
 """
 
+import os
 import sys
 from datetime import datetime
 from pathlib import Path
 
 from flask import Flask, render_template, send_from_directory
+
+from utils.debug_flag import resolve_debug_flag
 
 from api.workspaces import bp as workspaces_bp
 from api.composers import bp as composers_bp
@@ -101,6 +104,13 @@ if __name__ == "__main__":
         help="Path to exclusion rules file (sensitive projects/chats are omitted). "
              "If omitted, uses ~/.cursor-chat-browser/exclusion-rules.txt if present.",
     )
+    parser.add_argument(
+        "--debug",
+        action="store_true",
+        help="Enable Flask debug mode and the Werkzeug debugger. "
+             "DANGEROUS: allows remote code execution if the port is exposed. "
+             "Off by default; can also be enabled via FLASK_DEBUG=1.",
+    )
     args = parser.parse_args()
 
     if args.base_dir:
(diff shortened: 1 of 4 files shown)


tests/test_cli_args.py
```diff
@@ -246,5 +248,84 @@ def test_export_py_has_since_choices(self):
...
+        for node in ast.walk(tree):
+            if not isinstance(node, ast.Call):
+                continue
+            for kw in node.keywords:
+                if (
+                    kw.arg == "debug"
+                    and isinstance(kw.value, ast.Constant)
+                    and kw.value.value is True
+                ):
+                    offenders.append(kw.lineno)
```
bradjin8: The AST guard only flags kw.arg == "debug" with ast.Constant and value is True. Shapes like debug=True under Python 3.7 used ast.NameConstant (not Constant), so on very old interpreters this test could miss a literal. If you only support 3.8+ (or 3.11+ per CI), this is fine; otherwise extend the check to treat NameConstant with value is True like Constant.

tests/test_cli_args.py
bradjin8, line 1: Line 304-327

The guard targets literal keyword debug=True on a call. It does not catch semantically equivalent tricks (app.run(**{"debug": True}), debug=some_var where some_var is True). That matches the issue (“literal … re-introduced”); just know determined bypasses are still possible—acceptable for the stated goal.

app.py
```diff
@@ -109,10 +119,23 @@ def favicon():
 
     app = create_app(exclusion_rules_path=args.exclude_rules)
     print(f"Cursor Chat Browser (Python) running at http://{args.host}:{args.port}")
+
+    debug_enabled = resolve_debug_flag(os.environ.get("FLASK_DEBUG"), args.debug)
```
bradjin8, line 123: Only FLASK_DEBUG is consulted via resolve_debug_flag. Operators used to flask run might expect FLASK_ENV=development to imply debug; this code path does not read it. Not required by issue #9 (which names FLASK_DEBUG=1 only); worth a one-line README note if confusion shows up in support.

app.py
```diff
@@ -109,10 +119,23 @@ def favicon():
...
+    if debug_enabled:
+        # Print the warning to stderr so it's visible even when stdout is
+        # piped/redirected. The Werkzeug debugger is a remote-code-execution
+        # primitive — anyone reaching the host:port can hijack the process.
+        print(
+            "WARNING: Flask debug mode ENABLED. The Werkzeug debugger allows "
+            "arbitrary code execution by anyone who can reach this server. "
+            "Bind only to 127.0.0.1 and never expose to untrusted networks.",
+            file=sys.stderr,
+        )
```
bradjin8, line 133: Comments use Unicode em dashes (—), which can trigger GitHub’s “bidirectional / hidden Unicode” banner on the PR. Purely cosmetic; switch to ASCII - if you want a quiet diff view.

bradjin8 (changes requested):
## READ.md - mention for the `debug` is off by default.

PR/issue are security-sensitive; a short note that debug is off by default and enabled only with --debug or FLASK_DEBUG would match operator expectations. Optional; you asked not to expand scope unless useful.

bradjin8 (approved):

PR #21, opened 2026-05-07
fix: sanitise Marked.js output with DOMPurify (closes #11)

### LINKED ISSUE #11: Security: sanitise Marked.js output with DOMPurify (XSS via javascript: links / data: URIs)
## Problem

`marked.parse(...)` is called twice in the frontend and its output is injected straight into the DOM (or into a downloadable HTML file) without sanitisation:

- [`templates/workspace.html:317`](templates/workspace.html#L317) — `renderMarkdown(text)` returns `marked.parse(text, { breaks: true, gfm: true })`, which then flows into `main.innerHTML` at [line 248](templates/workspace.html#L248) via `bubble.text` rendering.
- [`static/js/download.js:191`](static/js/download.js#L191) — `marked.parse(md)` is interpolated directly into a downloadable `.html` blob.

`marked.parse` does not sanitise. Any chat content reaching this path can carry XSS payloads — most commonly via `javascript:` URLs in markdown links, dangerous `data:` URIs, `<script>` blocks (which Marked passes through), and event-handler attributes on whitelisted tags. While chat content *should* be operator-authored, the threat model includes:

- Shared `.cursor-chat` exports from third parties
- Future Cursor extensions / plugins writing into the local chat store
- Pasted content from external tools that retains attacker-controlled markdown
- Anyone who can write to the user's `~/.cursor/` workspace storage directory

A successful XSS in the dashboard origin can read every chat in the local store and exfiltrate it.

## Suggested fix

- Load **DOMPurify** from CDN in `templates/base.html` alongside Marked.js and highlight.js.
- Add a `renderMarkdownSafe(text)` helper in `static/js/app.js` that wraps `marked.parse(...)` with `DOMPurify.sanitize(...)`. If either library failed to load, fall back to `escapeHtml(text)` rather than emitting raw markdown HTML.
- Replace the two unsanitised call sites (`workspace.html`, `download.js`) with the new helper.
- Ship regression tests that fail if any future change reintroduces a bare `marked.parse(...)` flowing into `innerHTML` or a download blob without `DOMPurify.sanitize`.

## Severity

Critical — XSS in the dashboard origin reads the entire local chat store. Listed as a Critical / 5pt item in Will's eval week-1 plan for `cppa-cursor-browser`.

timon0305:
## Problem

`marked.parse(...)` doesn't sanitise. Two call sites — `templates/workspace.html` (→ `innerHTML`) and `static/js/download.js` (→ download blob) — were piping its output straight into the DOM, exposing XSS via `javascript:` links, `data:` URIs, raw `<script>` tags, and event-handler attributes. A hit reads the entire local chat store.

## Change

- `templates/base.html` — load DOMPurify 3.2.4 (after Marked.js, before `app.js`).
- `static/js/app.js` — new `renderMarkdownSafe(text)` helper: `marked.parse → DOMPurify.sanitize`. Falls back to `escapeHtml` if either lib is missing — never a bare Marked call.
- `templates/workspace.html` and `static/js/download.js` — both call sites now use the helper.

## Test plan

- 145/145 unit tests pass (8 new in `tests/test_xss_sanitization.py`, including a source-level guard that fails if any future frontend file outside the helper contains a bare `marked.parse(...)`).
- Live smoke: home returns 200, served HTML loads both Marked + DOMPurify, `app.js` exposes `renderMarkdownSafe`.
- Browser DevTools verification (run on a workspace page):
  - `renderMarkdownSafe('[click](javascript:alert(1))')` → `href` stripped
  - `renderMarkdownSafe('<script>alert(1)</script>')` → `<script>` stripped
  - `renderMarkdownSafe('# H\n\n**bold**')` → full HTML preserved (sanity)

Closes #11.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced markdown rendering safety across the UI, including sanitized HTML output for exported HTML files and consistent escaping of conversation metadata to reduce XSS risk.

* **Tests**
  * Added regression tests to ensure the sanitized rendering helper is used and rendering safety standards are preserved.

### DIFF
--- a/static/js/app.js
+++ b/static/js/app.js
@@ -48,6 +48,30 @@ function escapeHtml(str) {
   return div.innerHTML;
 }
 
+/**
+ * Render Markdown to HTML, then sanitise with DOMPurify before returning.
+ *
+ * Marked.js does NOT sanitise. Without DOMPurify, `[link](javascript:...)`,
+ * dangerous `data:` URIs, and inline event handlers all survive into the DOM
+ * — that's the XSS class fixed in issue #11.
+ *
+ * Fallback: if either marked or DOMPurify is missing (CDN failure, ad blocker,
+ * tests), return the plain-text-escaped string rather than ever emit raw or
+ * unsanitised HTML. Never fall through to a bare Marked call without sanitising.
+ */
+function renderMarkdownSafe(text) {
+  if (!text) return '';
+  if (typeof marked === 'undefined' || typeof DOMPurify === 'undefined') {
+    return escapeHtml(text);
+  }
+  try {
+    const html = marked.parse(text, { breaks: true, gfm: true });
+    return DOMPurify.sanitize(html);
+  } catch (e) {
+    return escapeHtml(text);
+  }
+}
+
 function formatDate(timestamp) {
   if (!timestamp) return '';
   try {
--- a/templates/base.html
+++ b/templates/base.html
@@ -9,10 +9,14 @@
   <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/11.9.0/styles/vs2015.min.css" id="hljs-theme">
   <script src="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/11.9.0/highlight.min.js"></script>
   <!-- Marked.js for Markdown rendering -->
-  <script src="https://cdnjs.cloudflare.com/ajax/libs/marked/12.0.1/marked.min.js"></script>
-  <!-- DOMPurify — sanitises Marked.js output before any DOM insertion (issue #11). -->
-  <!-- Required: every marked.parse(...) call must flow through DOMPurify.sanitize(...). -->
-  <script src="https://cdnjs.cloudflare.com/ajax/libs/dompurify/3.2.4/purify.min.js"></script>
+  <script src="https://cdnjs.cloudflare.com/ajax/libs/marked/12.0.1/marked.min.js"
+    integrity="sha384-3zf4Pen4fXU90jGg3cxmo7BF4dq8HMtF2s07c/Hhd1Fh+Encm6ApPvNm8vrMJbWu"
+    crossorigin="anonymous"></script>
+  <!-- DOMPurify - sanitises Marked.js output before any DOM insertion (issue #11). -->
+  <!-- Required: Marked output must be passed through DOMPurify.sanitize before DOM use. -->
+  <script src="https://cdnjs.cloudflare.com/ajax/libs/dompurify/3.2.4/purify.min.js"
+    integrity="sha384-eEu5CTj3qGvu9PdJuS+YlkNi7d2XxQROAFYOr59zgObtlcux1ae1Il3u7jvdCSWu"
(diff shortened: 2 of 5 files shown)


tests/test_xss_sanitization.py
```diff
@@ -0,0 +1,143 @@
...
+    frontend — the one inside the helper itself."""
+
+    FRONTEND_FILES = [
+        "templates/workspace.html",
+        "templates/index.html",
+        "templates/search.html",
+        "templates/config.html",
+        "static/js/app.js",
+        "static/js/download.js",
+    ]
```
bradjin8: FRONTEND_FILES is a fixed list. A new static/js/*.js or template that calls marked.parse would not be scanned until the list is updated. Optional hardening: discover static/js/*.js and templates/*.html automatically, or add a CI grep for marked.parse across the repo.

templates/base.html
```diff
@@ -10,6 +10,9 @@
   <script src="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/11.9.0/highlight.min.js"></script>
   <!-- Marked.js for Markdown rendering -->
   <script src="https://cdnjs.cloudflare.com/ajax/libs/marked/12.0.1/marked.min.js"></script>
+  <!-- DOMPurify — sanitises Marked.js output before any DOM insertion (issue #11). -->
+  <!-- Required: every marked.parse(...) call must flow through DOMPurify.sanitize(...). -->
+  <script src="https://cdnjs.cloudflare.com/ajax/libs/dompurify/3.2.4/purify.min.js"></script>
```
bradjin8: CDN scripts have no Subresource Integrity (SRI). Same pattern as Marked/highlight already; adding SRI hashes for Marked + DOMPurify would reduce CDN tampering risk (tradeoff: bump chore on upgrades).

Comments use Unicode (—); GitHub may show the “bidirectional Unicode” banner. Optional ASCII-only comments.

static/js/app.js
```diff
@@ -48,6 +48,30 @@ function escapeHtml(str) {
...
+ * unsanitised HTML. Never fall through to a bare Marked call without sanitising.
+ */
+function renderMarkdownSafe(text) {
+  if (!text) return '';
+  if (typeof marked === 'undefined' || typeof DOMPurify === 'undefined') {
+    return escapeHtml(text);
+  }
+  try {
+    const html = marked.parse(text, { breaks: true, gfm: true });
+    return DOMPurify.sanitize(html);
```
bradjin8, line 69: DOMPurify.sanitize(html) uses defaults. That is appropriate for XSS; if you later see legitimate Markdown features stripped, you may need an explicit config (still keeping dangerous URLs/scripts out). Not a defect for #11.

templates/workspace.html
bradjin8, line 1: Line 154-185, 196-217

Metadata strings (models, token lines, etc.) are interpolated into HTML without escapeHtml in some branches; issue #11 is specifically Markdown → HTML. Worth a separate pass if API-fed strings can contain &lt; / script-like content—orthogonal to Marked XSS.

bradjin8 (changes requested):

bradjin8 (approved):

PR #22, opened 2026-05-07
fix: validate /api/set-workspace path — reject traversal, symlink escape, non-existent (closes #15)

timon0305:
## Problem

`POST /api/set-workspace` accepted any string in `body.path`, did `~/` expansion, and stored the result in a module-global consumed by every subsequent file lookup. A client (anyone who reaches the dashboard, including a hostile JS payload — see related XSS issue #11) could post `{ "path": "/etc" }` or `{ "path": "/home/user/.ssh" }` and the app would then serve those files as Cursor data. Symlink escape (e.g. `/tmp/cursor-link → /`) bypassed any naive `startswith` check.

## Change

- **`utils/path_validation.py`** (new) — `validate_workspace_path(raw)` with `WorkspacePathError`. `os.path.realpath()` collapses `..` AND resolves symlinks in one step; rejects non-string/empty, non-existent, non-directory, no-Cursor-markers; returns the canonical real path so the override is stored canonically, not as the caller sent it. Lives outside `api/` so tests don't pull Flask into scope (existing test-suite convention).
- **`api/config_api.py`** — handler calls the validator. Returns `400 { error: "<reason>" }` on rejection (was silently `200`); on accept returns `200 { success: true, path: "<canonical>" }`.
- **`tests/test_workspace_path_validation.py`** (new) — 11 regression cases.

## Test plan

Unit (`python3 -m unittest discover tests`): **148/148 OK** (was 137; +11 new).

Live HTTP smoke against the running app, all 9 cases:

| Case | Payload | Expected | Got |
|------|---------|----------|-----|
| A | empty body | `400 path is required` | ✅ |
| B | empty path string | `400 path is required` | ✅ |
| C | non-existent path | `400 path does not exist` | ✅ |
| D | file `/etc/passwd` | `400 path is not a directory` | ✅ |
| E | dir `/etc` (no markers) | `400 no Cursor workspaceStorage` | ✅ |
| F | traversal `…/storage/../..` lands at `/tmp` | `400` (`..` collapsed, then markers reject) | ✅ |
| G | symlink → `/` | `400` (realpath unmasked it) | ✅ |
| H | real fake workspace (`state.vscdb` in subdir) | `200 { path: "<canonical>", success: true }` | ✅ |
| I | symlink → real workspace | `200 { path: "<realpath>" }` (canonicalised — input was the symlink path) | ✅ |

Case I is the meaningful one: even though the operator sent `…/good-link`, the override stored is the realpath. Every subsequent read goes through canonical, not the link — a future swap of the link target can't redirect reads.

Closes #15.

## Summary by CodeRabbit

* **Improvements**
  * Stricter workspace path validation: rejects invalid/traversal inputs and non-workspace locations, resolves canonical real paths, and returns canonical path plus workspace count on validation.
  * Setting a workspace now validates input and returns the canonical path on success; malformed/non-object JSON yields clear 400 errors.
* **Tests**
  * Added regression tests covering validation, canonicalization, symlink behavior, and API responses.
* **Documentation**
  * README clarified workspace-path validation and operator escape hatch.
* **Bug Fixes**
  * UI shows server-provided validation error messages.

### DIFF
--- a/api/config_api.py
+++ b/api/config_api.py
@@ -53,6 +53,10 @@ def validate_path():
     """Same path rules as POST /api/set-workspace: realpath, markers (issue #15)."""
     try:
         body = request.get_json(silent=True) or {}
+        if not isinstance(body, dict):
+            return jsonify(
+                {"valid": False, "error": "invalid JSON body", "workspaceCount": 0}
+            )
         raw = body.get("path", "")
         try:
             canonical = validate_workspace_path(raw)
--- a/api/config_api.py
+++ b/api/config_api.py
@@ -76,7 +76,14 @@ def validate_path():
 
 @bp.route("/api/set-workspace", methods=["POST"])
 def set_workspace():
-    body = request.get_json(silent=True) or {}
+    # Reject non-dict JSON bodies (array / string / number / null). Without
+    # this, get_json returns the value directly, the truthy fallback `or {}`
+    # is bypassed, and `body.get("path", "")` raises AttributeError — which
+    # the outer Exception handler then mis-reports as a 500 server error
+    # instead of a 400 client error. (CodeRabbit on PR #16.)
+    body = request.get_json(silent=True)
+    if not isinstance(body, dict):
+        return jsonify({"error": "request body must be a JSON object"}), 400
     raw = body.get("path", "")
     # Validate the supplied path BEFORE storing the override (issue #15).
     # validate_workspace_path collapses `..` traversal AND resolves symlinks
--- a/api/config_api.py
+++ b/api/config_api.py
@@ -12,7 +12,6 @@ import sys
 
 from flask import Blueprint, jsonify, request
 
-from utils.path_helpers import expand_tilde_path
 from utils.path_validation import WorkspacePathError, validate_workspace_path
 from utils.workspace_path import set_workspace_path_override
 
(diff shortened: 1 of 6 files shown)


tests/test_workspace_path_validation.py
```diff
@@ -0,0 +1,201 @@
...
+    @unittest.skipIf(sys.platform == "win32", "POSIX symlinks only")
+    def test_symlink_to_real_workspace_is_canonicalised_and_accepted(self):
+        # Symlink → real Cursor storage. Accepted, but the canonical path
+        # returned is the realpath (the storage dir), NOT the symlink path.
+        storage = _make_cursor_workspace_dir(self.tmp)
+        link = os.path.join(self.tmp, "good-link")
+        os.symlink(storage, link)
+        result = validate_workspace_path(link)
+        self.assertEqual(result, os.path.realpath(storage))
+        self.assertNotEqual(result, link)
```
bradjin8, line 142: Symlink tests skipped on Windows — Linux/macOS CI still exercises them; fine if CI matrix includes non-Windows.

utils/path_validation.py
```diff
@@ -0,0 +1,83 @@
...
+    expanded = expand_tilde_path(raw_path)
+    # realpath() collapses `..` AND resolves symlinks. Both classes of escape
+    # become equivalent to whatever is actually on disk.
+    real = os.path.realpath(expanded)
+
+    if not os.path.exists(real):
+        raise WorkspacePathError("path does not exist")
+    if not os.path.isdir(real):
+        raise WorkspacePathError("path is not a directory")
+    if not _has_cursor_workspace_markers(real):
```
bradjin8, line 82: Classic TOCTOU: path can change between realpath and listdir. Hard to fix fully without broader locking; low practical risk for this app.

api/config_api.py
bradjin8, line 1: /api/validate-path still uses expand_tilde_path only (no realpath), unlike set-workspace. Symlinks / .. can make validate-path and set-workspace disagree. Issue #15 scoped set-workspace; optional follow-up is to reuse validate_workspace_path (or shared realpath + markers) here so the UI and API stay aligned.

bradjin8 (changes requested):
## utils/workspace_path.py 61–67 
WORKSPACE_PATH env is still applied in resolve_workspace_path() without the same checks as the API. That is a normal “trusted operator” escape hatch but differs from XSS-driven POST abuse in #15. Optional: document in README, or apply the same validator when the env is read (stricter).

## utils/path_validation.py 30–48
Marker heuristic only inspects immediate children for state.vscdb. Matches /api/validate-path and the issue; if Cursor ever nests storage differently, both endpoints could false-negative. Acceptable as stated.

bradjin8 (approved):

PR #23, opened 2026-05-07
fix: wrap all production sqlite3.connect() in context managers (closes #17)

timon0305:
## Problem

19 production `sqlite3.connect()` call sites in this repo used a bare local variable with no `with` block, no `try/finally`, and no guaranteed `.close()` on exception. On a long-lived Flask process serving many requests, leaks accumulate before the user notices (file descriptors, SQLite read-locks held longer than needed).

## Change

Two patterns, each applied where it fits best:

**`with closing(sqlite3.connect(...)) as conn:` — short-scope (15 sites)**
- `api/composers.py` ×3
- `utils/cli_chat_reader.py` ×2
- `utils/cursor_md_exporter.py` ×1
- `api/workspaces.py` ×2 (small helpers)
- `api/logs.py` ×2
- `api/search.py` ×2 (`wconn` + small `conn`)
- `api/export_api.py` ×1 (`wconn`)

`sqlite3.Connection`'s own context manager only commits / rolls back; it does not `.close()`. `contextlib.closing(...)` is what guarantees release.

**`try / except / finally: if conn is not None: conn.close()` — long-scope (4 sites)**
- `api/workspaces.py:get_workspace_tabs` (~400 lines using `global_db`)
- `api/search.py:search` ×2 deep blocks (~160-line and ~70-line scopes)
- `api/export_api.py:export_chats` (~340 lines)
- `scripts/export.py:_conn` (cleaned up the existing duplicate close-on-success and close-on-error pattern)

Same semantics as wrapping the whole function in `with closing(...)`, without the 100-300-line indent shift.

**`@contextmanager` on the helper that yields a connection — 1 site**

`api/workspaces.py:_open_global_db` previously returned `(conn, path)` to callers, who were responsible for `.close()`. Converted to a `@contextmanager` so callers write `with _open_global_db(...) as (conn, _):`. Both call sites updated.

Test-side `sqlite3.connect` calls are out of scope (per-test fixture lifecycles, short-lived).

## Test plan

- `pytest tests/` → **137 / 137 OK** (zero behaviour change in the test paths).
- AST parses cleanly on all 8 modified files; `from contextlib import closing[, contextmanager]` correctly added everywhere it's used.
- Static guard: every `sqlite3.connect()` call site classified and confirmed protected by either `closing()`, an outer `try/finally` block in the same function, or `@contextmanager`.
- Live HTTP smoke against the running app:
  - `GET /` → HTTP 200
  - `GET /api/workspaces` → HTTP 200 (exercises `with closing()` + `@contextmanager` + small `with closing()` in `_build_composer_id_to_workspace_id`)
  - `GET /api/logs` → HTTP 200 (both `with closing()` paths)
  - `GET /api/composers` → HTTP 200 (all 3 `with closing()` sites)
  - `GET /api/search?q=test&type=all` → HTTP 200 (both `try/finally` sites + small `with closing()`)
  - No Python tracebacks in the server log.

Closes #17.

## Summary by CodeRabbit

* **Chores**
  * Improved database connection handling across APIs, export flows, logs, search, workspace listing, and CLI utilities to ensure deterministic cleanup and reduce resource leaks.
* **Documentation**
  * Added a "Tests" section to README with commands to run the Python test suite.
* **Tests**
  * Updated a test module to run via Python's unittest command.

### DIFF
--- a/api/workspaces.py
+++ b/api/workspaces.py
@@ -11,6 +11,7 @@ import json
 import os
 import re
 import sqlite3
+from pathlib import Path
 import sys
 from contextlib import closing, contextmanager
 from datetime import datetime, timezone
@@ -160,7 +161,10 @@ def _infer_workspace_name_from_context(workspace_path: str, workspace_id: str) -
     composer_ids: list[str] = []
     try:
         # closing() guarantees .close() on scope exit (issue #17).
-        with closing(sqlite3.connect(f"file:{local_db_path}?mode=ro", uri=True)) as lconn:
+        # Path.as_uri() percent-encodes reserved chars (#, ?, spaces, etc.);
+        # naive f"file:{path}" breaks sqlite URI parsing.
+        _db_uri = Path(local_db_path).resolve().as_uri() + "?mode=ro"
+        with closing(sqlite3.connect(_db_uri, uri=True)) as lconn:
             row = lconn.execute(
                 "SELECT value FROM ItemTable WHERE [key] = 'composer.composerData'"
             ).fetchone()
--- a/api/workspaces.py
+++ b/api/workspaces.py
@@ -953,15 +953,10 @@ def get_workspace_tabs(workspace_id):
     if workspace_id.startswith("cli:"):
         return _get_cli_workspace_tabs(workspace_id)
 
-    # `global_db` is the long-lived connection used across the ~400-line body
-    # below. Wrap the whole function in try/finally so the connection is
-    # released on every exit path — including unexpected exceptions that
-    # bypass the in-body close calls (issue #17). Equivalent to wrapping the
-    # body in `with closing(sqlite3.connect(...))` without the indent shift.
-    global_db = None
+    # Global DB reads use `_open_global_db` (issue #17) — same lifecycle as
+    # list_workspaces; connection closes when the `with` block exits.
     try:
         workspace_path = resolve_workspace_path()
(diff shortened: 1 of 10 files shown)


api/workspaces.py
bradjin8, line 1: ## 508–513 vs 1009–1015

Two patterns for long-lived connections: _open_global_db uses yield + finally, get_workspace_tabs uses **global_db = None + outer finally. Both correct; optional future refactor to reuse _open_global_db for get_workspace_tabs would dedupe (not required for #17).

bradjin8, line 1: 253 - 257

except Exception swallows errors in the global-storage search block but finally still closes conn — behavior preserved; logging-only downside pre-existed.

scripts/export.py
```diff
@@ -396,11 +396,11 @@ def main():
...
-            _conn.close()
-            _conn = None
         except Exception as e:
             print(f"Warning: Could not read Cursor IDE chats ({e}) — skipping.", file=sys.stderr)
+        finally:
+            # Guaranteed close on every exit path (issue #17). Replaces the
+            # previous duplicate close-in-success-and-error pattern.
             if _conn is not None:
                 try:
                     _conn.close()
```
bradjin8, line 406: Line 401-408: 
finally wraps close() in try/except so close failures do not mask earlier errors — sensible.

bradjin8 (changes requested):

bradjin8 (approved):

wpak-ai (approved):

Issue #25, opened 2026-05-10
Monolith extraction: split api/workspaces.py first module


timon0305:
## Problem

`api/workspaces.py` is 1,408 LOC and co-locates database access, multi-stage dict transformation, workspace resolution, project-assignment heuristics, CLI session handling, and HTTP response formatting in a single file. The `get_workspace_tabs` route handler alone spans lines 956-1407 (~450 lines). `_determine_project_for_conversation` (lines 269-390) contains 120+ lines of cascading fallback heuristics. Helpers are duplicated across files: `_read_json_file` is defined identically in `api/workspaces.py:66` and `api/composers.py:20`. Any change to workspace logic risks regressing conversation rendering.

## Severity

High · 8 points

## Acceptance criteria

- [ ] DB helpers (`_open_global_db`, `_collect_workspace_entries`, `_build_composer_id_to_workspace_id`, `_collect_invalid_workspace_ids`) extracted to `services/workspace_db.py`
- [ ] Resolution logic (`_determine_project_for_conversation`, `_infer_invalid_workspace_aliases`, `_create_project_name_to_workspace_id_map`, `_create_workspace_path_to_id_map`, `_get_workspace_display_name`, `_infer_workspace_name_from_context`) extracted to `services/workspace_resolver.py`
- [ ] CLI workspace tab logic (`_get_cli_workspace_tabs`) extracted to `services/cli_tabs.py`
- [ ] `_read_json_file` deduplicated to a single shared location
- [ ] `api/workspaces.py` route handlers contain only request parsing, service calls, and response formatting — no direct SQLite access or dict transformation logic (target <300 LOC, down from 1,408)
- [ ] All existing tests pass without modification (behavior-preserving refactor; JSON shapes unchanged)
- [ ] CI green
- [ ] PR approved by at least 1 reviewer

Issue #1, opened 2026-02-18
Add an optional parameter to disable rendering/exporting sensitive projects/chats


wpak-ai:
## Description

This new optional parameter will be accompanied with the path (either relative to the current working directory or absolute path) of an exclusion rule file that supports filtering out all projects and chats that contain the following combinations (anywhere in project/chat title, model names, chats, etc) from browser rendered results or exported files:
- a combination of keywords with unlimited number of AND or OR operators
- double-quoted phrases for exact matches

## Deliverables:
- an optional command-line parameter that points to the above-mentioned exclusion rule file
- works in browser mode as well as in CLI mode (`app.py` and `export.py`)

### The exclusion rule file should be 

- easy to understand and use, and hard to misinterpret or misuse
- in textual format that supports all locales and languages

### The behavior when not specified

- if there is this file in `.cursor-chat-browser` subdirectory of user's home directory (come up with an appropriate name and extension if any), it will always be used
- otherwise, no filtering happens

### Properly test whether the exclusion rule file works or not

PR #30, opened 2026-05-11
feat: typed models + schema validation at DB read boundaries (closes #24)

timon0305:
Closes #24.

## What

Adds a typed-model layer at the four Cursor SQLite read sites. Each `from_dict()` raises `SchemaError` when a critical field is missing, so a schema rename surfaces as a printed warning instead of an empty conversation rendered silently.

## Why

Per the eval (Section 5.1 — *Schema Fragility*), Cursor's on-disk JSON shapes are not versioned. Every read site used `.get("field")` with a fallback, so a Cursor rename of `composerData`, `bubbleId`, `latestRootBlobId`, etc. would return `None`, the fallback would apply, and the user would see an empty conversation with no error in the log. The eval called this *"the tool's primary failure mode, unobservable by any automated means."*

## How

**New `models/` package** — six dataclasses, each with a `from_dict()` validator:

| Model | Used at |
|---|---|
| `Composer` | global `cursorDiskKV` composer rows |
| `WorkspaceLocalComposer` | per-workspace `ItemTable` composer summaries |
| `Bubble` | `bubbleId:*` rows |
| `CliSessionMeta` | CLI session `meta` blob |
| `Workspace` | `workspace.json` |
| `ExportEntry` | export manifest (`manifest.jsonl`) |

`SchemaError` inherits from `ValueError` so existing `except ValueError` call sites still trap it.

**Wire-ins** (each catches `SchemaError`, prints `Schema drift in ...`, skips the row):

| Read site | Model |
|---|---|
| `api/workspaces.py:601` | `Composer` |
| `api/composers.py:57-77` | `WorkspaceLocalComposer` (per-row) + envelope dict/list guards |
| `api/search.py:163-164` | `Composer` |
| `utils/cli_chat_reader.py:93-98` | `CliSessionMeta`, with broadened catch for bad-hex / non-UTF-8 / bad-JSON |

## Verification

| Check | Result |
|---|---|
| `python3 -m unittest discover tests -v` | **204 pass** (178 existing + 26 new) |
| `mypy --no-strict-optional models/` | `Success: no issues found in 6 source files` |
| Live smoke against my Cursor workspaceStorage (17 composers) | All routes 200; **zero `Schema drift` warnings** — no false positives on real data |

## Notes

- `createdAt` is gated as required on `Composer` per the issue spec. A live scan confirmed 17/17 real composers carry it; one in-repo test fixture (`test_search_exclusion_filtering`) was updated to match.
- `_read_json_file` was duplicated in `api/workspaces.py:66` and `api/composers.py:20`. **Still duplicated** on this branch (consolidation lives in PR #31 — the `api/workspaces.py` split for #25).
- Eight follow-up commits in this PR address CodeRabbit findings: non-dict guards on every `from_dict`, drop of `or []` and `str()` coercion fallbacks that masked drift, Symbol-throw safety in catch arms, structured-error contract on per-row workspace-local composer drift, and `except Exception: pass` → `print` on the outer read loop.

## Summary by CodeRabbit

* **Refactor**
  * Stronger schema-aware validation and normalization for workspaces, composers, bubbles, and conversations across read paths.

* **Bug Fixes**
  * Malformed or drifted records are logged and skipped; search/workspace/composer results use validated headers and timestamps; CLI meta errors return empty results safely.

* **New Features**
  * New typed models and package-level exports for composer, bubble, workspace, export entry, CLI session meta, and schema errors.

* **Chores**
  * Export manifest now includes title and workspace and skips invalid lines.

* **Tests**
  * Added extensive model, read-path, export, and blob-traversal regression tests.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cppalliance/cppa-cursor-browser/pull/30)

### DIFF
--- a/api/composers.py
+++ b/api/composers.py
@@ -13,6 +13,7 @@ from flask import Blueprint, jsonify
 
 from utils.workspace_path import resolve_workspace_path
 from utils.path_helpers import to_epoch_ms
+from models import SchemaError
 
 bp = Blueprint("composers", __name__)
 
@@ -54,13 +55,24 @@ def list_composers():
 
                 if row and row[0]:
                     data = json.loads(row[0])
+                    if "allComposers" not in data:
+                        raise SchemaError("WorkspaceComposers", "allComposers")
                     all_composers = data.get("allComposers")
-                    if isinstance(all_composers, list):
-                        for c in all_composers:
-                            c["conversation"] = c.get("conversation") or []
-                            c["workspaceId"] = name
-                            c["workspaceFolder"] = workspace_folder
-                            composers.append(c)
+                    if not isinstance(all_composers, list):
+                        raise SchemaError(
+                            "WorkspaceComposers",
+                            "allComposers",
+                            hint=f"expected list, got {type(all_composers).__name__}",
+                        )
+                    for c in all_composers:
+                        if not isinstance(c, dict) or not c.get("composerId"):
+                            continue
+                        c["conversation"] = c.get("conversation") or []
+                        c["workspaceId"] = name
+                        c["workspaceFolder"] = workspace_folder
+                        composers.append(c)
+            except SchemaError as e:
+                print(f"Schema drift in {db_path}: {e}")
             except Exception:
                 pass
 
--- a/api/composers.py
+++ b/api/composers.py
@@ -13,7 +13,7 @@ from flask import Blueprint, jsonify
 
 from utils.workspace_path import resolve_workspace_path
 from utils.path_helpers import to_epoch_ms
-from models import SchemaError, WorkspaceLocalComposer
+from models import Composer, SchemaError, Workspace, WorkspaceLocalComposer
 
 bp = Blueprint("composers", __name__)
 
@@ -41,9 +41,11 @@ def list_composers():
 
             workspace_folder = None
             try:
-                wd = _read_json_file(wj_path)
-                workspace_folder = wd.get("folder")
-            except Exception:
+                workspace = Workspace.from_dict(_read_json_file(wj_path), workspace_id=name)
+                workspace_folder = workspace.folder
+            except (SchemaError, OSError, ValueError):
+                # Missing / malformed workspace.json is non-fatal — the row still
+                # contributes its composer data, just without a folder hint.
                 pass
 
             try:
@@ -72,21 +74,27 @@ def list_composers():
                         )
                     for c in all_composers:
                         try:
-                            WorkspaceLocalComposer.from_dict(c)
+                            local = WorkspaceLocalComposer.from_dict(c)
                         except SchemaError as e:
                             print(f"Schema drift in {db_path}: {e}")
                             continue
+                        # Use the typed view downstream so the dataclass is
+                        # load-bearing, not just a filter (Brad's review): the
+                        # sort key and the JSON's composerId both read off the
+                        # validated values, not the raw dict.
+                        c["composerId"] = local.composer_id
+                        c["lastUpdatedAt"] = local.last_updated_at
                         c["conversation"] = c.get("conversation") or []
                         c["workspaceId"] = name
                         c["workspaceFolder"] = workspace_folder
-                        composers.append(c)
+                        composers.append((local, c))
             except SchemaError as e:
                 print(f"Schema drift in {db_path}: {e}")
             except Exception as e:
                 print(f"Failed reading composers from {db_path}: {e}")
 
-        composers.sort(key=lambda c: to_epoch_ms(c.get("lastUpdatedAt")), reverse=True)
-        return jsonify(composers)
+        composers.sort(key=lambda pair: to_epoch_ms(pair[0].last_updated_at), reverse=True)
+        return jsonify([c for _, c in composers])
 
     except Exception as e:
         print(f"Failed to get composers: {e}")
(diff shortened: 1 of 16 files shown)


api/workspaces.py
bradjin8, line 1: ## lines 1097–1116 — get_workspace_tabs bypasses Composer.from_dict in the same file that validates at line 604

list_workspaces() validates every composerData row with Composer.from_dict (line 604) and skips drift with a warning. get_workspace_tabs() in the same module still uses raw json.loads + cd.get("fullConversationHeadersOnly") or []. A composer missing createdAt or with a malformed fullConversationHeadersOnly is hidden from the workspace list but can still render in tabs (or vice versa). This undercuts the issue's goal — schema drift remains silent on one of the two primary conversation-browsing paths.

api/composers.py
bradjin8, line 1: ##  lines 96–142: get_composer has no schema validation

list_composers validates each row via WorkspaceLocalComposer.from_dict and logs drift. get_composer returns the raw dict from either per-workspace DB or global storage with no validation. A drifted single-composer fetch returns malformed JSON to the client with no warning — exactly the silent-failure mode issue #24 targets.

models/errors.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class SchemaError(ValueError):
+    """Raised when a required Cursor schema field is missing or malformed."""
+
+    def __init__(self, model: str, field: str, *, hint: str | None = None) -> None:
+        self.model = model
+        self.field = field
+        self.hint = hint
+        message = f"{model}: missing required field '{field}'"
```
bradjin8: Type mismatches (e.g. createdAt is a string, fullConversationHeadersOnly is a dict) also raise SchemaError but the message always reads "missing required field". Operators grepping logs for drift will misread type errors as absent keys. Consider "invalid field" when hint is present, or separate message templates.

api/composers.py
```diff
@@ -54,15 +55,35 @@ def list_composers():
...
+                    for c in all_composers:
+                        try:
+                            WorkspaceLocalComposer.from_dict(c)
+                        except SchemaError as e:
+                            print(f"Schema drift in {db_path}: {e}")
+                            continue
+                        c["conversation"] = c.get("conversation") or []
+                        c["workspaceId"] = name
+                        c["workspaceFolder"] = workspace_folder
+                        composers.append(c)
```
bradjin8: ## validation result is discarded

WorkspaceLocalComposer.from_dict(c) is used purely as a gate; the typed object is thrown away and the raw mutating dict is appended. Fine for a phase-1 wire-in, but the typed layer provides no actual type safety downstream — only a filter. Same pattern could use the returned WorkspaceLocalComposer for composer_id / last_updated_at to prove the model is load-bearing.

models/conversation.py
```diff
@@ -0,0 +1,122 @@
...
+        if "createdAt" not in raw:
+            raise SchemaError("Composer", "createdAt")
+
+        created_at = raw.get("createdAt")
+        if not isinstance(created_at, (int, float)) or isinstance(created_at, bool):
+            raise SchemaError(
+                "Composer",
+                "createdAt",
+                hint=f"expected timestamp number, got {type(created_at).__name__}",
+            )
```
bradjin8, line 50: ## createdAt must be int/float, rejects ISO strings

The PR documents a live scan showing 17/17 real composers use numeric timestamps, so this is a deliberate choice — not a bug on current data. Worth noting as a potential false positive if Cursor ever stores ISO-8601 strings in createdAt; those composers would disappear from list/search with a drift warning while tabs (issue #1) might still show them.

bradjin8 (changes requested):
## False Positives / Correctly Handled
- WorkspaceLocalComposer vs Composer split — Correct design. Per-workspace rows lack fullConversationHeadersOnly / createdAt; using Composer.from_dict there would reject every real record. CodeRabbit's round-2 suggestion to use Composer.from_dict(c) was rightly rejected and replaced with WorkspaceLocalComposer.
- createdAt required on Composer — Aligned with issue #24 after the round-1 fix; fixtures updated; test_missing_created_at_raises replaces the old tolerant test.
- CodeRabbit rounds 1–3 — All actionable findings appear addressed in follow-up commits (non-dict guards, dropped or [] / str() coercion, broadened CLI catch, per-row SchemaError routing, ID type gates, createdAt numeric gate).
- SchemaError(ValueError) — Preserves compatibility with existing except ValueError sites; tested in test_schema_error_inherits_value_error.

api/workspaces.py
bradjin8: This is the most important finding: half of `api/workspaces.py`'s composer traffic is validated and half is not, in the same PR that claims to close the schema-fragility issue. I'd request that before merge, at minimum.

timon0305: Fixed. `get_workspace_tabs()` now uses `Composer.from_dict` with the same SchemaError catch as `list_workspaces()` — both paths agree on what counts as a valid composer. 
Added a regression test so this can't regress quietly: it patches the composer model and asserts the tabs endpoint actually calls `from_dict`.

api/composers.py
timon0305: Fixed. `get_composer()` now validates both paths the same way `list_composers()` does — per-workspace entries go through `WorkspaceLocalComposer.from_dict` (drift logged + skipped), and the global-fallback path goes through `Composer.from_dict` (drift returns 404 instead of leaking malformed JSON). Two regression tests pin both branches.

models/errors.py
timon0305: Done. `SchemaError` now reads `"missing required field 'X'"` only when the field is genuinely absent (no hint), and `"invalid field 'X' (expected …, got …)"` when there's a type/shape mismatch. Added a test that asserts the two wordings can't cross over.

api/composers.py
timon0305: Good call. The typed object is now load-bearing — captured as `local`, used to overwrite `composerId` and `lastUpdatedAt` on the output dict, and the sort key reads `local.last_updated_at` instead of the raw dict. Added a regression test that patches `from_dict` to return a different `last_updated_at` than the raw dict and asserts the sort follows the typed value, so if anyone reverts to raw-dict access it fails loudly.

models/conversation.py
timon0305: Noted. Added a 4-line WHY comment at the `isinstance(created_at, (int, float))` check pointing at the 17/17 numeric scan, so a future contributor reading the strict gate doesn't "fix" it without that context. Leaving the behaviour as-is until Cursor actually starts emitting ISO strings — at which point we relax, not silently coerce.

bradjin8 (approved):

wpak-ai (approved):

PR #31, opened 2026-05-12
refactor(api): split api/workspaces.py 1,407 → 142 LOC into services/ (closes #25)

timon0305:
Closes #25.

## What

Splits `api/workspaces.py` from **1,407 → 142 LOC** into focused service modules. Three thin route handlers remain in `api/workspaces.py`; everything else moves to `services/` and `utils/workspace_descriptor.py`. Behaviour-preserving — every existing test passes without modification.

## Why

Per the eval (Section 5.2 — *Monolith Handler*), `api/workspaces.py` was a 1,400+ LOC monolith co-locating SQLite reads, dict transformation, workspace resolution, project-assignment heuristics, CLI session handling, and HTTP response formatting in one file. `get_workspace_tabs` alone was ~450 lines. `_read_json_file` was duplicated identically in `api/composers.py:20`. The file was resistant to unit testing and a single change to workspace logic risked regressing conversation rendering.

## How

**Extraction layout** (in dependency order — pure helpers first, route bodies last):

| New module | Functions moved from `api/workspaces.py` |
|---|---|
| `utils/workspace_descriptor.py` | `_basename_from_pathish`, `_resolve_workspace_descriptor`, `_uri_or_path_to_fs_path`, `_read_json_file` (dedup from `api/composers.py:20`) |
| `services/workspace_db.py` | `_collect_workspace_entries`, `_collect_invalid_workspace_ids`, `_build_composer_id_to_workspace_id`, `_open_global_db` |
| `services/workspace_resolver.py` | `_get_workspace_display_name`, `_infer_workspace_name_from_context`, `_get_project_from_file_path`, `_create_project_name_to_workspace_id_map`, `_create_workspace_path_to_id_map`, `_determine_project_for_conversation`, `_infer_invalid_workspace_aliases` |
| `services/cli_tabs.py` | `_get_cli_workspace_tabs` |
| `services/workspace_listing.py` | body of `GET /api/workspaces` (`list_workspace_projects(...)`) |
| `services/workspace_tabs.py` | body of `GET /api/workspaces/<id>/tabs` (`assemble_workspace_tabs(...)`) |

**`api/workspaces.py` is now 142 LOC**: three route handlers that resolve the request, hand off to a service, and JSONify the result.

[...]

PR #32, opened 2026-05-13
feat(tests): pytest endpoint coverage via Flask test client (closes #26)

timon0305:
Closes #26.

## What

Adds pytest integration tests for the four conversation-browsing HTTP routes (`/api/workspaces`, `/api/workspaces/<id>`, `/api/workspaces/<id>/tabs`, `/api/search`) via Flask's `app.test_client()`. Both `unittest discover` and the new `pytest tests/` are CI merge gates.

## Why

Per the eval (§5.2), the four HTTP routes had **zero** automated coverage — existing tests covered utility functions only. JSON-shape regressions were undetectable without a human in the loop.

## How

**`tests/conftest.py`** — three pytest fixtures:

| Fixture | What |
|---|---|
| `workspace_storage` | Temp layout with seeded `globalStorage/state.vscdb` + per-workspace `state.vscdb` + `workspace.json`. Sets `WORKSPACE_PATH` and `CLI_CHATS_PATH` envs; restores on cleanup. |
| `client` | Flask test client bound to the seeded storage. |
| `empty_workspace_client` | Flask test client over an empty workspaceStorage — for 404 paths. |

**`tests/test_api_endpoints.py`** — 11 tests:

| Endpoint | Happy | 404 / empty | Edge |
|---|:---:|:---:|:---:|
| `GET /api/workspaces` | ✅ | ✅ empty → `[]` | — |
| `GET /api/workspaces/<id>` | ✅ | ✅ unknown → 404 | ✅ `global` → "Other chats" |
| `GET /api/workspaces/<id>/tabs` | ✅ | ✅ missing global storage → 404 | ✅ `global` → tabs |
| `GET /api/search?q=...` | ✅ finds sentinel | ✅ no-match → `{"results": []}` | ✅ missing `q` tolerated (200 or 400) |

**`.github/workflows/tests.yml`** — `pytest>=8` added to install, new "Run pytest integration suite" step after the unittest step.

## Verification

| Check | Result |
|---|---|
| `python -m unittest discover tests -v` | 178 pass (no regression) |
| `python -m pytest tests/ -v --tb=short` | 189 pass (178 + 11 new) |
| mypy on new files | zero new errors (4 errors are pre-existing on master) |
| Production code modified | **none** — `api/`, `services/`, `utils/`, `models/` untouched |

## Notes

- This PR is **test infrastructure only**. The Flask routes are exercised against fixture data, not modified.
- Orthogonal to PR #31 (the `api/workspaces.py` split) — can merge in either order.

## Summary by CodeRabbit

* **Tests**
  * Added a comprehensive integration test suite for API endpoints (workspaces list/details, tabs, search) covering seeded vs empty storage, error cases (404/400), exclusion-rule behavior, and helpers/fixtures to bootstrap isolated test workspaces and shared test IDs.

* **Chores**
  * CI updated to run the pytest-based integration tests as a separate merge gate alongside existing unit tests.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cppalliance/cppa-cursor-browser/pull/32)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -53,9 +53,11 @@ jobs:
 
       - name: Run pytest integration suite
         # Pytest fixtures (tests/conftest.py) build a temp workspaceStorage
-        # and exercise the Flask routes via app.test_client(). Runs alongside
-        # unittest, not instead of — both suites are merge gates.
-        run: python -m pytest tests/ -v --tb=short
+        # and exercise the Flask routes via app.test_client(). Scoped to the
+        # new endpoint file because `pytest tests/` would also re-collect the
+        # 178 unittest.TestCase subclasses already run in the step above —
+        # ~2× the CI minutes for zero extra signal.
+        run: python -m pytest tests/test_api_endpoints.py -v --tb=short
 
   # ── Typecheck: mypy ───────────────────────────────────────────────────────
   # Codebase already has type hints across most of the surface (~70+ typed
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -52,9 +52,9 @@ jobs:
         run: python -m unittest discover tests -v
 
       - name: Run pytest integration suite
-        # Closes #26. Pytest fixtures (tests/conftest.py) build a temp
-        # workspaceStorage and exercise the Flask routes via app.test_client().
-        # Runs alongside unittest, not instead of — both suites are merge gates.
+        # Pytest fixtures (tests/conftest.py) build a temp workspaceStorage
+        # and exercise the Flask routes via app.test_client(). Runs alongside
+        # unittest, not instead of — both suites are merge gates.
         run: python -m pytest tests/ -v --tb=short
 
   # ── Typecheck: mypy ───────────────────────────────────────────────────────
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -46,10 +46,16 @@ jobs:
         # system packages on Linux — out of scope for the unittest suite.
         run: |
           python -m pip install --upgrade pip
-          python -m pip install 'flask>=3.0' 'fpdf2>=2.7'
+          python -m pip install 'flask>=3.0' 'fpdf2>=2.7' 'pytest>=8'
 
       - name: Run unittest suite
         run: python -m unittest discover tests -v
+
+      - name: Run pytest integration suite
+        # Closes #26. Pytest fixtures (tests/conftest.py) build a temp
+        # workspaceStorage and exercise the Flask routes via app.test_client().
+        # Runs alongside unittest, not instead of — both suites are merge gates.
+        run: python -m pytest tests/ -v --tb=short
 
   # ── Typecheck: mypy ───────────────────────────────────────────────────────
   # Codebase already has type hints across most of the surface (~70+ typed
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,161 @@
+from __future__ import annotations
+
+import json
+import os
+import sqlite3
+import sys
+import tempfile
+from pathlib import Path
+from typing import Generator
+
+import pytest
+
+REPO_ROOT = str(Path(__file__).resolve().parent.parent)
+if REPO_ROOT not in sys.path:
+    sys.path.insert(0, REPO_ROOT)
+
+from app import create_app
+
+
+HAPPY_COMPOSER_ID = "cmp-happy"
+HAPPY_BUBBLE_ID = "bub-happy"
+HAPPY_WORKSPACE_ID = "ws-happy"
+
+
+def _make_global_state_db(path: str) -> None:
+    """globalStorage/state.vscdb with one composerData + one bubbleId row."""
+    conn = sqlite3.connect(path)
+    conn.execute("CREATE TABLE cursorDiskKV ([key] TEXT PRIMARY KEY, value TEXT)")
+    conn.execute(
+        "INSERT INTO cursorDiskKV ([key], value) VALUES (?, ?)",
+        (
+            f"composerData:{HAPPY_COMPOSER_ID}",
+            json.dumps({
+                "name": "Happy conversation",
+                "createdAt": 1_715_000_000_000,
+                "lastUpdatedAt": 1_715_000_500_000,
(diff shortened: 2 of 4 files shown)


.github/workflows/tests.yml
```diff
@@ -46,11 +46,17 @@ jobs:
...
+          python -m pip install 'flask>=3.0' 'fpdf2>=2.7' 'pytest>=8'
 
       - name: Run unittest suite
         run: python -m unittest discover tests -v
 
+      - name: Run pytest integration suite
+        # Pytest fixtures (tests/conftest.py) build a temp workspaceStorage
+        # and exercise the Flask routes via app.test_client(). Runs alongside
+        # unittest, not instead of — both suites are merge gates.
+        run: python -m pytest tests/ -v --tb=short
```
bradjin8: ## pytest re-runs all 178 existing unittest tests, doubling CI time (unresolved CodeRabbit finding)

Because pytest collects unittest.TestCase subclasses, pytest tests/ runs the 178 existing tests a second time on top of the python -m unittest discover step. Each matrix cell executes 178 + 189 = 367 tests; across 3 Python versions that is 1,101 total test executions vs the original 534 — roughly double the CI minutes with no new signal. The intended gate is only the 13 new integration tests. 

Fix: scope the step to python -m pytest tests/test_api_endpoints.py -v --tb=short.

tests/test_api_endpoints.py
```diff
@@ -0,0 +1,130 @@
+from __future__ import annotations
+
+from tests.conftest import HAPPY_BUBBLE_ID, HAPPY_COMPOSER_ID, HAPPY_WORKSPACE_ID
```
bradjin8: ## importing constants directly from conftest is fragile (unresolved CodeRabbit finding)

conftest.py is special to pytest — under non-default import modes (--import-mode=importlib, used by many modern pytest configs) it may not be importable as tests.conftest. The constants should live in a regular helper module (e.g., tests/test_ids.py) and be imported from there by both conftest.py and test_api_endpoints.py.

tests/conftest.py
```diff
@@ -0,0 +1,161 @@
...
+            f"bubbleId:{HAPPY_COMPOSER_ID}:{HAPPY_BUBBLE_ID}",
+            json.dumps({
+                "text": "find me by search term sentinel-grep",
+                "type": "user",
+                "createdAt": 1_715_000_400_000,
+            }),
+        ),
+    )
+    conn.commit()
+    conn.close()
```
bradjin8: ##  DB helpers don't guard conn.close() against mid-setup exceptions

If any conn.execute() call raises (e.g., disk full, unexpected schema conflict), conn.close() is never reached and the connection leaks, potentially locking the tempdir file and blocking subsequent cleanup. Use contextlib.closing or wrap in try/finally. Same pattern in _make_workspace. In practice this is low risk in a temp environment, but it's a fixture quality issue.

tests/conftest.py
```diff
@@ -0,0 +1,161 @@
...
+    conn.execute("CREATE TABLE ItemTable ([key] TEXT PRIMARY KEY, value TEXT)")
+    conn.execute(
+        "INSERT INTO ItemTable ([key], value) VALUES (?, ?)",
+        (
+            "composer.composerData",
+            json.dumps({"allComposers": [{"composerId": HAPPY_COMPOSER_ID}]}),
+        ),
+    )
+    conn.commit()
+    conn.close()
```
bradjin8: ##  DB helpers don't guard conn.close() against mid-setup exceptions

If any conn.execute() call raises (e.g., disk full, unexpected schema conflict), conn.close() is never reached and the connection leaks, potentially locking the tempdir file and blocking subsequent cleanup. Use contextlib.closing or wrap in try/finally. Same pattern in _make_workspace. In practice this is low risk in a temp environment, but it's a fixture quality issue.

tests/test_api_endpoints.py
```diff
@@ -0,0 +1,130 @@
...
+        # The seeded user bubble must be present
+        bubble_types = [b["type"] for b in tab["bubbles"]]
+        assert "user" in bubble_types
+
+    def test_global_returns_tabs(self, client):
+        response = client.get("/api/workspaces/global/tabs")
+        assert response.status_code == 200
+        body = response.get_json()
+        assert "tabs" in body and isinstance(body["tabs"], list)
+
```
bradjin8, line 96: ## test_global_returns_tabs is too weak to catch isolation bugs

The seeded HAPPY_COMPOSER_ID is assigned to HAPPY_WORKSPACE_ID, so global tabs should be empty []. The test passes with tabs == [], which is correct, but it doesn't assert that the global bucket is actually empty (i.e., that the workspace-assignment logic properly isolated the seeded conversation). Adding assert HAPPY_COMPOSER_ID not in [t["id"] for t in body["tabs"]] would turn this into a real isolation check.

tests/test_api_endpoints.py
bradjin8, line 1: ## No test exercises exclusion rules

EXCLUSION_RULES is hardcoded to [] in both fixtures. Issue #26 requires endpoint coverage, but the exclusion-rule path (present on /api/workspaces, /api/workspaces/<id>/tabs, /api/search) has no integration test. A simple case — seed a workspace whose name matches an exclusion pattern and assert it is absent from the list — would close this gap. Not a blocker for merge but a coverage hole.

tests/conftest.py
```diff
@@ -0,0 +1,161 @@
...
+def client(workspace_storage: str):
+    """Flask test client bound to the temp workspace_storage fixture."""
+    app = create_app()
+    app.config["TESTING"] = True
+    app.config["EXCLUSION_RULES"] = []
+    return app.test_client()
+
+
+@pytest.fixture
+def empty_workspace_client() -> Generator:
```
bradjin8: ## Missing type parameters on Generator return annotation

workspace_storage correctly annotates Generator[str, None, None]. empty_workspace_client uses bare Generator with no type parameters, which mypy treats as Generator[Any, Any, Any]. Should be Generator[Flask test client type, None, None] or simply drop the annotation and let pytest infer it. Minor but inconsistent with the adjacent fixture.

bradjin8 (changes requested):

bradjin8 (approved):

wpak-ai (approved):

PR #33, opened 2026-05-13
test: CLI export end-to-end + manual QA checklist (closes #27)

timon0305:
Closes #27.

## Summary

- **`tests/test_cli_export_e2e.py`** — five unittest cases that drive `scripts/export.py` `main()` against a self-contained SQLite fixture (workspaceStorage + globalStorage `cursorDiskKV`) under `tempfile`, with `WORKSPACE_PATH` / `CLI_CHATS_PATH` / `XDG_STATE_HOME` overrides. The fixture is local to the file so the branch does not depend on PR #32's `conftest.py` being merged first; the suite runs on master CI as-is under `unittest discover`.
- **`tests/cli-export-qa-checklist.md`** — manual companion checklist covering `--help`, zip + no-zip exports, `--since last` idempotency, `python app.py --port 3001` launch + `curl` smoke, and the browse-flow markdown/PDF downloads.
- **`.gitignore`** — adds `exports/` next to the existing `export/` entry so either pluralisation of the example output directory stays untracked.

### Automated coverage

| Test | Asserts |
|---|---|
| `test_export_no_zip_writes_markdown_files` | `--no-zip` produces `.md` files; seeded composer id appears in a filename |
| `test_markdown_frontmatter_has_required_fields` | frontmatter contains `log_id`, `title`, `workspace`, `created_at`, `updated_at` |
| `test_export_zip_mode_writes_archive` | default mode writes `cursor-export-<date>.zip` containing `.md` entries |
| `test_manifest_jsonl_has_expected_shape` | `manifest.jsonl` entries carry `log_id`, `path`, `updated_at` |
| `test_since_last_skips_already_exported_records` | second `--since last` run leaves existing files' mtimes untouched |

The `--since last` test wraps the inner `main()` call in a `SystemExit` guard because, on a clean second pass, the script prints `No conversations found since last export.` and exits 0 — the no-rewrite contract is still verifiable through the unchanged file mtimes.

## Test plan

- [x] `python3 -m unittest tests.test_cli_export_e2e -v` — 5/5 pass locally
- [x] `python3 -m unittest discover tests` — 183/183 pass, no regression
- [x] `python3 app.py --port 3001` boots without traceback; `curl -sI http://127.0.0.1:3001/` returns `HTTP/1.1 200 OK`
- [ ] CI green (matrix: ubuntu-latest × Python 3.11 / 3.12 / 3.13)
- [ ] Manual QA per `tests/cli-export-qa-checklist.md` on a real Cursor profile

## Summary by CodeRabbit

* **New Features**
  * CLI: support for using the XDG state directory and an option to skip IDE composer exports (--no-composer).

* **Tests**
  * Added end-to-end tests validating CLI export: markdown/zip output, manifest creation, incremental exports, and exclusion rules.

* **Documentation**
  * Added a QA checklist for CLI export and web UI verification steps.

* **Chores**
  * Updated ignore rules to exclude generated export artifacts.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cppalliance/cppa-cursor-browser/pull/33)

### DIFF
--- a/tests/cli-export-qa-checklist.md
+++ b/tests/cli-export-qa-checklist.md
@@ -68,7 +68,7 @@ python app.py
 ```
 
 - [ ] Process stays running for at least 30 s without crash
-- [ ] `curl -sI http://127.0.0.1:3000/` returns `HTTP/1.0 200 OK` (or 200 over 1.1)
+- [ ] `curl -sI http://127.0.0.1:3000/` returns `HTTP/1.1 200 OK` (or `HTTP/1.0 200 OK` under some dev-server configurations)
 - [ ] Home page lists at least one workspace card (assuming real Cursor data)
 - [ ] No `Exception` / `Traceback` in server log
 
--- a/tests/cli-export-qa-checklist.md
+++ b/tests/cli-export-qa-checklist.md
@@ -24,7 +24,7 @@ Mark each row pass/fail and attach short notes for any unexpected output.
 
 ## 2. CLI: default zip export
 
-```
+```bash
 python scripts/export.py --out ./export
 ```
 
@@ -37,7 +37,7 @@ python scripts/export.py --out ./export
 
 ## 3. CLI: no-zip export
 
-```
+```bash
 python scripts/export.py --out ./export --no-zip
 ```
 
@@ -49,7 +49,7 @@ python scripts/export.py --out ./export --no-zip
 
 ## 4. CLI: incremental (`--since last`)
 
-```
+```bash
 python scripts/export.py --out ./export --no-zip --since last
 ```
 
@@ -63,18 +63,18 @@ python scripts/export.py --out ./export --no-zip --since last
 
 ## 5. App server launch
 
-```
-python app.py --port 3001
+```bash
+python app.py
 ```
 
 - [ ] Process stays running for at least 30 s without crash
-- [ ] `curl -sI http://127.0.0.1:3001/` returns `HTTP/1.0 200 OK` (or 200 over 1.1)
+- [ ] `curl -sI http://127.0.0.1:3000/` returns `HTTP/1.0 200 OK` (or 200 over 1.1)
 - [ ] Home page lists at least one workspace card (assuming real Cursor data)
 - [ ] No `Exception` / `Traceback` in server log
 
 ## 6. Browse flow (web UI)
 
-Open `http://127.0.0.1:3001/` in a browser.
+Open `http://127.0.0.1:3000/` in a browser.
 
 - [ ] Workspace list renders without console errors
 - [ ] Clicking a workspace card opens the workspace view
--- a/tests/cli-export-qa-checklist.md
+++ b/tests/cli-export-qa-checklist.md
@@ -1,0 +1,90 @@
+# CLI export + browse — manual QA checklist
+
+Companion to the automated `tests/test_cli_export_e2e.py` suite. Run this once
+per release branch (or whenever `scripts/export.py`, `app.py`, or the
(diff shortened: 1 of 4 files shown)


tests/test_cli_export_e2e.py
bradjin8, line 1: ## lines 28–57 and 73–80 — conn.close() not guarded

Same pattern as PR #32's conftest helpers: _make_global_state_db and _make_workspace_storage call conn.close() at the end of the function without try/finally. Any exception mid-setup leaks the open connection and can lock the temp SQLite file, causing TemporaryDirectory.cleanup() to fail on Windows (which refuses to delete open files). Wrap in try/finally: conn.close() or use with closing(sqlite3.connect(...)) as conn:.

bradjin8, line 1: ## No coverage for --no-composer or --exclude-rules flags

The export script has two behavior-changing flags with non-trivial logic: --no-composer (skips the entire IDE composer block) and --exclude-rules (filters by project/title). Neither has a test. For --no-composer in particular, the test fixture seeds IDE composer data exclusively — a --no-composer run should produce zero output, which is easy to verify. These are gaps for a "CLI testing" issue, though not blockers.

tests/cli-export-qa-checklist.md
```diff
@@ -0,0 +1,90 @@
...
+      only the new file
+
+## 5. App server launch
+
+```bash
+python app.py
+```
+
+- [ ] Process stays running for at least 30 s without crash
+- [ ] `curl -sI http://127.0.0.1:3000/` returns `HTTP/1.0 200 OK` (or 200 over 1.1)
```
bradjin8: ## HTTP/1.0 listed first, likely to confuse testers

Werkzeug's dev server responds HTTP/1.0 when the client sends HTTP/1.0, but curl -sI sends HTTP/1.1 by default and typically gets HTTP/1.1 200 OK. Listing HTTP/1.0 first is likely to trigger false-fail reports from testers who see HTTP/1.1. Suggest reversing the phrasing: HTTP/1.1 200 OK (or HTTP/1.0 200 OK in some dev-server configurations).

bradjin8 (changes requested):
## tests/test_cli_export_e2e.py, line 97 and scripts/export.py, lines 131–133: --since last test is not hermetically isolated


_run_export sets XDG_STATE_HOME to a temp directory with the clear intent of redirecting where export_state.json is saved and read. But scripts/export.py's get_global_state_dir() hardcodes ~/.cursor-chat-browser and never reads XDG_STATE_HOME. The state file is therefore always written to the real user's home directory.

The --since last test still passes on a clean machine because the sequence is:

First run (no --since) writes ~/.cursor-chat-browser/export_state.json with lastExportTime = now (2026).
Second run (--since last) reads it back, finds lastExportTime (2026) > fixture.lastUpdatedAt (2024), skips the composer, exits 0.
SystemExit(0) is caught; mtime check passes since files weren't rewritten.
The test passes by coincidence of the timestamp gap, not by design. The side effects are real:

On any developer machine that runs the test suite, ~/.cursor-chat-browser/export_state.json gets overwritten with lastExportTime = now. Their next scripts/export.py --since last run will skip all conversations with timestamps before the test ran.
tearDown only cleans up self._tmp; ~/.cursor-chat-browser is never restored.

Fix: make get_global_state_dir() in scripts/export.py honor XDG_STATE_HOME when set (standard XDG convention). Alternatively, mock get_global_state_dir in the test with unittest.mock.patch.

bradjin8 (approved):

wpak-ai (approved):

PR #51, opened 2026-05-18
Refactor: Deduplicate normalize_file_path and to_epoch_ms in export script (closes #46)

bradjin8:
Closes #46

## Problem

`scripts/export.py` maintained private copies of `normalize_file_path` and `to_epoch_ms` that already existed in `utils/path_helpers.py`. The two `normalize_file_path` copies had diverged: the export copy was missing the non-Windows cross-platform branch that handles Windows-style drive paths (`D:/...`, `D:\...`) on Linux/macOS — meaning cross-platform reads of Cursor's Windows workspaceStorage could produce inconsistent path keys depending on which code path was used.

## Changes

**`scripts/export.py`**
- Removed local `normalize_file_path` (~13 lines) and `to_epoch_ms` (~22 lines)
- Expanded the `utils.path_helpers` import to include both functions alongside the existing `get_workspace_folder_paths` alias

**`tests/test_normalize_file_path.py`** (new)
- 21 test cases in this file (12 `normalize_file_path`, 9 `to_epoch_ms`; 2 skip on win32 for POSIX passthrough only) covering the shared implementation:
  - `file:///` and `file://` URI stripping
  - Percent-encoded characters: `%20` (space), `%23` (#), `%3A` (colon in URI-style prefix)
  - Windows drive paths with backslashes and forward-slashes on all platforms
  - Drive-letter lowercasing
  - Plain POSIX path passthrough
  - Empty string

## Notes

- The canonical implementation in `utils/path_helpers.py` is a strict superset of the old export copy — it adds the non-Windows `^[a-zA-Z]:[/\\]` branch for Windows-style paths on Linux/macOS, which the export copy lacked.
- The `sys.path.insert` block in `scripts/export.py` is intentionally left in place; removing it is part of the larger export-script-to-service-layer refactor (issue #1).
- `samples/example_chat_export.md` quotes the old function definitions as documentation; not touched.

## Verification

```
python -m unittest tests.test_normalize_file_path -v
# Ran 21 tests … OK (skipped=2)   # win32: 2 POSIX-only cases skip

python -m unittest discover tests -q
# Ran 274 tests in 2.1s
# OK (skipped=4)   # suite-wide; includes skips from other modules
```

## Summary by CodeRabbit

* **Refactor**
  * Consolidated utility functions into shared modules.

* **Tests**
  * Significantly expanded test coverage for path normalization and timestamp conversion utilities. Added comprehensive test cases covering cross-platform compatibility, percent-decoding, URI handling, and various timestamp input formats including ISO-8601 and numeric values.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cppalliance/cppa-cursor-browser/pull/51?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/tests/test_normalize_file_path.py
+++ b/tests/test_normalize_file_path.py
@@ -1,22 +1,20 @@
-"""Tests for utils.path_helpers.normalize_file_path.
+"""Tests for utils.path_helpers path/timestamp helpers (closes #46).
 
-Covers the shared implementation that was previously duplicated in
-scripts/export.py (closes #46). All call-sites in both the web app and the
-CLI export script now use this single copy.
+Covers ``normalize_file_path`` and ``to_epoch_ms``, both previously duplicated
+in scripts/export.py. All call-sites in the web app and CLI export script now
+use the shared implementations in utils.path_helpers.
 
-Edge-case matrix:
-  - file:/// and file:// URI schemes
-  - Percent-encoded characters: spaces (%20), colons (%3A), hashes (%23)
-  - Windows-style drive paths (backslash and forward-slash) on all platforms
-  - Drive-letter lowercasing on win32
-  - Plain POSIX paths pass through unchanged
-  - Empty / None-like input
+Test inventory (this module only): 21 cases — 12 ``normalize_file_path``,
+9 ``to_epoch_ms``. On win32, 2 cases skip (POSIX passthrough in
+``TestNormalizeFilePathPosixPassthrough`` only). A full-suite run may report
+more skips (e.g. ``skipped=4``) from other test modules, not this file.
 """
 
 import sys
 import unittest
+from datetime import datetime, timezone
 
-from utils.path_helpers import normalize_file_path
+from utils.path_helpers import normalize_file_path, to_epoch_ms
 
 
 class TestNormalizeFilePathUriStripping(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


tests/test_normalize_file_path.py
timon0305, line 1: The PR deduplicates BOTH functions but only adds tests for normalize_file_path. Grep across the repo shows there are zero existing tests for to_epoch_ms either (grep -rln "to_epoch_ms" tests/ returns nothing). So the function now has two call sites (export script + wherever else utils.path_helpers is used) with zero coverage.

tests/test_normalize_file_path.py
```diff
@@ -0,0 +1,107 @@
...
+        self.assertIn("my documents", out)
+
+    def test_hash_decoded(self) -> None:
+        out = normalize_file_path("file:///C:/repo/src%23internal/mod.py")
+        self.assertNotIn("%23", out)
+        self.assertIn("#", out)
+
+    def test_percent_encoded_colon_in_uri_prefix(self) -> None:
+        """URI-style /d%3A/... path: %3A is decoded to ':'.
+
```
timon0305, line 48: the comment block explaining the win32 vs non-win32 divergence is the kind of test-as-spec that helps future readers. One nit: this test is the only place the new cross-platform branch's behavior is exercised end-to-end with a leading /, so worth flagging as such in the docstring

tests/test_normalize_file_path.py
```diff
@@ -0,0 +1,107 @@
...
+    def test_mixed_case_drive_lowercased(self) -> None:
+        out = normalize_file_path(r"E:\Mixed\Case\Path")
+        self.assertTrue(out.startswith("e:"))
+        self.assertEqual(out, r"e:\mixed\case\path")
+
+
+class TestNormalizeFilePathPosixPassthrough(unittest.TestCase):
+    def test_plain_posix_path_unchanged_on_non_windows(self) -> None:
+        if sys.platform == "win32":
```
timon0305, line 96: Four tests, all check both platforms produce the same lowercased-backslash output. Solid. test_file_uri_with_windows_drive (line 87) is more permissive than the others — uses assertIn rather than assertEqual, presumably to tolerate platform-specific separator differences. Consider tightening to assertEqual once you've nailed down the expected output on each platform; otherwise a future regression that changes the trailing format could pass.

tests/test_normalize_file_path.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def test_path_without_scheme_unchanged(self) -> None:
+        if sys.platform == "win32":
+            self.skipTest("plain relative path behaviour differs on win32")
+        out = normalize_file_path("relative/path/file.py")
+        self.assertEqual(out, "relative/path/file.py")
+
+
+if __name__ == "__main__":
+    unittest.main()
```
timon0305, line 141: Both tests are skipped on win32. That's correct (POSIX semantics differ there) but the PR description says "12 test cases (4 platform-skipped)" — actually 2 are skipped on win32, not 4. The "skipped=4" in the verification block is the suite-wide skip count, not this file's. Minor doc accuracy nit.

timon0305 (changes requested):

tests/test_normalize_file_path.py
bradjin8: Added TestToEpochMs alongside the existing normalize_file_path tests — same file, shared utils.path_helpers coverage for both deduplicated functions.

timon0305 (approved):

wpak-ai (approved):

PR #52, opened 2026-05-18
Fix: Skip NULL bubble rows in workspace tabs loader (closes #50)

bradjin8:
Closes #50 

## Problem

`GET /api/workspaces/<id>/tabs` returned a 500 for any workspace that had a `cursorDiskKV` row with a `NULL` value column. The bubble-loading loop in `services/workspace_tabs.py` called `json.loads(row["value"])` without a `None` guard, and `json.loads(None)` raises `TypeError`. The existing `except (json.JSONDecodeError, ValueError)` clause did not catch `TypeError`, so the exception propagated all the way up and the project detail view was completely inaccessible.

## Root Cause

```python
# Before — TypeError on NULL value not caught
bubble_obj = Bubble.from_dict(json.loads(row["value"]), bubble_id=bid)
except (json.JSONDecodeError, ValueError):
    pass
```

## Fix

`services/workspace_tabs.py` — one guard added before `json.loads`:

```python
if row["value"] is None:
    continue
```

NULL rows are silently skipped; all other rows (valid, malformed JSON, schema drift) continue through their existing handlers unchanged.

## Tests

New file `tests/test_workspace_tabs_null_bubble.py` using a real SQLite temp DB:

- **`test_null_bubble_row_is_skipped_without_exception`** — workspace containing a NULL-value bubble row returns 200 with no `TypeError`
- **`test_healthy_bubbles_still_load_when_null_row_present`** — valid bubble rows alongside the NULL row are not dropped

## Verification

```
python -m unittest discover tests -q
# Ran 264 tests … OK (skipped=2)
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a crash when assembling workspace tabs by skipping null or unparseable stored entries; tabs now return successfully and continue loading remaining valid content.

* **Tests**
  * Added regression tests verifying null storage rows are ignored without error and that valid tabs, bubbles, and code diffs still load correctly.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cppalliance/cppa-cursor-browser/pull/52?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/services/workspace_tabs.py
+++ b/services/workspace_tabs.py
@@ -37,6 +37,16 @@ def _extract_chat_id_from_code_block_diff_key(key: str) -> str | None:
     return m.group(1) if m else None
 
 
+def _loads_disk_kv_value(raw: Any) -> Any | None:
+    """Parse a cursorDiskKV ``value`` column; ``None`` if missing or unparseable."""
+    if raw is None:
+        return None
+    try:
+        return json.loads(raw)
+    except (json.JSONDecodeError, TypeError, ValueError):
+        return None
+
+
 def assemble_workspace_tabs(
     workspace_id: str,
     workspace_path: str,
@@ -97,32 +107,30 @@ def assemble_workspace_tabs(
             parts = row["key"].split(":")
             if len(parts) >= 3:
                 bid = parts[2]
-                if row["value"] is None:
+                parsed = _loads_disk_kv_value(row["value"])
+                if parsed is None:
                     continue
                 try:
-                    bubble_obj = Bubble.from_dict(json.loads(row["value"]), bubble_id=bid)
+                    bubble_obj = Bubble.from_dict(parsed, bubble_id=bid)
                     bubble_map[bid] = bubble_obj.raw
                 except SchemaError as e:
                     # Drift logged so the operator can chase disappearing
                     # bubbles instead of guessing. Bad row still skipped so the
                     # tabs endpoint can't 500 on one malformed bubble.
                     print(f"Schema drift in bubble {bid}: {e}")
-                except (json.JSONDecodeError, ValueError):
-                    pass
 
         # Load codeBlockDiffs
         for row in _safe_fetchall("SELECT key, value FROM cursorDiskKV WHERE key LIKE 'codeBlockDiff:%'"):
             chat_id = _extract_chat_id_from_code_block_diff_key(row["key"])
             if not chat_id:
                 continue
-            try:
-                d = json.loads(row["value"])
-                code_block_diff_map.setdefault(chat_id, []).append({
-                    **d,
-                    "diffId": row["key"].split(":")[2] if len(row["key"].split(":")) > 2 else None,
-                })
-            except Exception:
-                pass
+            d = _loads_disk_kv_value(row["value"])
+            if not isinstance(d, dict):
+                continue
+            code_block_diff_map.setdefault(chat_id, []).append({
+                **d,
+                "diffId": row["key"].split(":")[2] if len(row["key"].split(":")) > 2 else None,
+            })
 
         # Load messageRequestContext rows once; build both
         # message_request_context_map and project_layouts_map from the same pass.
@@ -132,10 +140,7 @@ def assemble_workspace_tabs(
             if len(parts) < 2:
                 continue
             chat_id = parts[1]
-            try:
-                ctx = json.loads(row["value"])
-            except Exception:
-                continue
+            ctx = _loads_disk_kv_value(row["value"])
             if not isinstance(ctx, dict):
                 continue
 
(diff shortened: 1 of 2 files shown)


services/workspace_tabs.py
```diff
@@ -97,6 +97,8 @@ def _safe_fetchall(query: str, params: tuple = ()) -> list:
             parts = row["key"].split(":")
             if len(parts) >= 3:
                 bid = parts[2]
+                if row["value"] is None:
```
timon0305: bid = parts[2] / if row["value"] is None: continue
Correct fix in isolation. But this is the fourth distinct error-handling pattern for json.loads(row["value"]) in this single file:

Line 101 (bubble, this PR): pre-check is None + except (JSONDecodeError, ValueError)
Line 117 (codeBlockDiff): bare except Exception
Line 134 (messageRequestContext): bare except Exception
Line 155 (projectLayouts): bare except Exception
Line 182 (composer): except (JSONDecodeError, TypeError, ValueError)
The composer loader at line 182 already includes TypeError in its narrow except — that single change at line 108 would have caught this bug too, with no new line. Pick one pattern and apply it across all five sites.

tests/test_workspace_tabs_null_bubble.py
```diff
@@ -0,0 +1,109 @@
...
+        conn.close()
+
+        self.workspace_path = ws_dir
+
+    def tearDown(self):
+        self.tmp.cleanup()
+
+    def test_null_bubble_row_is_skipped_without_exception(self):
+        """assemble_workspace_tabs must not raise when a bubble row has NULL value."""
+        from services.workspace_tabs import assemble_workspace_tabs
```
timon0305: Import inside the test method instead of at module top. If deliberate (lazy DB init), worth a one-line comment. Otherwise hoist to module-level imports.

tests/test_workspace_tabs_null_bubble.py
```diff
@@ -0,0 +1,109 @@
...
+                rules=[],
+            )
+        except TypeError as exc:
+            self.fail(f"NULL bubble row raised TypeError: {exc}")
+
+        self.assertEqual(status, 200)
+
+    def test_healthy_bubbles_still_load_when_null_row_present(self):
+        """The healthy bubble surfaces in a tab even when a NULL row is present."""
+        from services.workspace_tabs import assemble_workspace_tabs
```
timon0305: Duplicate of line 67. Single module-level import would replace both.

tests/test_workspace_tabs_null_bubble.py
```diff
@@ -0,0 +1,109 @@
...
+            workspace_path=self.workspace_path,
+            rules=[],
+        )
+        self.assertEqual(status, 200)
+        self.assertIsInstance(payload, dict)
+        tabs = payload.get("tabs", [])
+        self.assertEqual(len(tabs), 1, "Expected exactly one tab for composer-abc")
+
+        tab = tabs[0]
+        self.assertEqual(tab["id"], "composer-abc")
```
timon0305: Assumes the composer id surfaces without its composerData: prefix. Worth a sanity check that this matches what the production endpoint returns — the temp DB will pass while real data shape drifts.

tests/test_workspace_tabs_null_bubble.py
```diff
@@ -0,0 +1,109 @@
...
+        self.assertEqual(tab["title"], "Test Chat")
+        self.assertIn("bubbles", tab)
+        self.assertIn("codeBlockDiffs", tab)
+
+        bubbles = tab["bubbles"]
+        self.assertEqual(len(bubbles), 1, "Expected exactly one bubble (null row skipped)")
```
timon0305, line 106: Good — assertion-with-message tells the reader what the test is proving. Same style would help lines 89, 90, 95, 96, 97, 98, 103, 104, 105.

timon0305 (changes requested):

tests/test_workspace_tabs_null_bubble.py
bradjin8: Moved from services.workspace_tabs import assemble_workspace_tabs to the top with the other imports (matches test_workspace_display_name.py and the rest of the suite). No lazy-init or circular-import reason to keep it inside the methods.

tests/test_workspace_tabs_null_bubble.py
bradjin8: Resolved — hoisted to line 15; both duplicate in-method imports removed.

tests/test_workspace_tabs_null_bubble.py
bradjin8: added assertion messages across both tests.

services/workspace_tabs.py
```diff
@@ -37,6 +37,16 @@ def _extract_chat_id_from_code_block_diff_key(key: str) -> str | None:
     return m.group(1) if m else None
 
 
+def _loads_disk_kv_value(raw: Any) -> Any | None:
```
timon0305: raw is typed Any but it's always a SQLite column value — narrower as raw: str | None would document the contract better. Minor.

timon0305: _loads_disk_kv_value is fine, but the silent-skip-on-error semantic isn't obvious from the name. Something like _safe_loads_kv_value or _try_loads_kv_value would signal "this returns None on failure rather than raising". Even smaller nit.

timon0305 (approved):

wpak-ai (approved):

PR #54, opened 2026-05-19
Fix thread-unsafe _workspace_path_override race (#43)

bradjin8:
## Summary

Fixes [#43](https://github.com/cppalliance/cppa-cursor-browser/issues/43) by serializing access to the module-level `_workspace_path_override` with a `threading.Lock`. The override is written by `POST /api/set-workspace` (and `--base-dir`) and read on every call to `resolve_workspace_path()`; without synchronization, threaded WSGI deployments (gunicorn `--threads`, waitress, etc.) can race and serve data from the wrong workspace path.

- Add `_workspace_path_lock` and document the locking contract on `_workspace_path_override`
- Protect reads in `get_workspace_path_override()` and `resolve_workspace_path()` (snapshot under lock, then expand outside)
- Protect writes in `set_workspace_path_override()`; widen type to `str | None` for test/cleanup clears
- Add `tests/test_workspace_path_thread_safety.py` with concurrent set/resolve and clear/set/resolve stress tests

Eval reference: `workspace-path-race-condition` cluster (test 13) in the [April 2026 Python eval](https://github.com/cppalliance/cppa-cursor-browser) baseline.

## Why a lock (not Flask `g`/session)

The override is **server-wide** configuration for this localhost tool, not per-request state. A lock is the minimum change that makes threaded deployment safe without changing API semantics. Per-user isolation would require session-scoped storage and is out of scope for this fix.

## Test plan

- [x] `python -m unittest tests.test_workspace_path_thread_safety -v`
- [x] `python -m unittest tests.test_workspace_path_validation -v` (existing set-workspace regressions still pass)
- [ ] Manual: run app under a threaded server, call `POST /api/set-workspace` while hitting list/search endpoints concurrently; confirm stable workspace selection

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a concurrency issue so workspace path overrides remain consistent and never produce mixed or partial results during concurrent set/clear/resolve operations.
  * Clear/set races now yield stable, predictable resolved paths.

* **Tests**
  * Added concurrency regression tests that validate stable behavior of workspace path overrides under concurrent set, clear, and resolve scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cppalliance/cppa-cursor-browser/pull/54?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/tests/test_workspace_path_thread_safety.py
+++ b/tests/test_workspace_path_thread_safety.py
@@ -1,0 +1,124 @@
+"""
+Regression tests for issue #43 — thread-safe _workspace_path_override.
+
+Run:
+    python -m unittest tests.test_workspace_path_thread_safety -v
+"""
+
+from __future__ import annotations
+
+import os
+import shutil
+import sys
+import tempfile
+import threading
+import unittest
+from concurrent.futures import ThreadPoolExecutor, as_completed
+
+REPO_ROOT = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+sys.path.insert(0, REPO_ROOT)
+
+from utils.workspace_path import (
+    get_workspace_path_override,
+    resolve_workspace_path,
+    set_workspace_path_override,
+)
+
+
+class TestWorkspacePathThreadSafety(unittest.TestCase):
+    """Concurrent set-workspace + resolve must not observe torn global state."""
+
+    def setUp(self):
+        self.tmp = tempfile.mkdtemp(prefix="cursor-ws-thread-test-")
+        self.addCleanup(shutil.rmtree, self.tmp, ignore_errors=True)
+        self.path_a = os.path.join(self.tmp, "storage-a")
+        self.path_b = os.path.join(self.tmp, "storage-b")
+        os.makedirs(self.path_a)
+        os.makedirs(self.path_b)
+        self._prior_workspace_env = os.environ.pop("WORKSPACE_PATH", None)
+        self.addCleanup(self._restore_workspace_env)
+        self.addCleanup(set_workspace_path_override, None)
+
+    def _restore_workspace_env(self):
+        if self._prior_workspace_env is None:
+            os.environ.pop("WORKSPACE_PATH", None)
+        else:
+            os.environ["WORKSPACE_PATH"] = self._prior_workspace_env
+
+    def test_concurrent_set_and_resolve_never_returns_mixed_paths(self):
+        iterations = 500
+        errors: list[str] = []
+        start = threading.Barrier(9)  # 1 writer + 8 readers
+        # Seed before workers start so readers never observe the unset default path.
+        set_workspace_path_override(self.path_a)
+
+        def writer() -> None:
+            start.wait()
+            for i in range(iterations):
+                set_workspace_path_override(self.path_a if i % 2 == 0 else self.path_b)
+
+        def reader() -> None:
+            start.wait()
+            for _ in range(iterations):
+                override = get_workspace_path_override()
+                if override is None:
+                    errors.append("override was unexpectedly cleared during run")
+                    continue
+                if override not in (self.path_a, self.path_b):
+                    errors.append(f"override returned unexpected value: {override!r}")
+                    continue
+                resolved = resolve_workspace_path()
+                expected = os.path.realpath(override)
+                if resolved != expected:
+                    errors.append(
+                        f"resolve {resolved!r} != realpath(override) {expected!r}"
+                    )
+
+        with ThreadPoolExecutor(max_workers=9) as pool:
+            futures = [pool.submit(writer)]
+            futures.extend(pool.submit(reader) for _ in range(8))
+            for fut in as_completed(futures):
+                fut.result()
+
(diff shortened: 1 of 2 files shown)


tests/test_workspace_path_thread_safety.py
```diff
@@ -0,0 +1,118 @@
...
+        self.tmp = tempfile.mkdtemp(prefix="cursor-ws-thread-test-")
+        self.addCleanup(shutil.rmtree, self.tmp, ignore_errors=True)
+        self.path_a = os.path.join(self.tmp, "storage-a")
+        self.path_b = os.path.join(self.tmp, "storage-b")
+        os.makedirs(self.path_a)
+        os.makedirs(self.path_b)
+        self.allowed_resolved = {
+            os.path.realpath(self.path_a),
+            os.path.realpath(self.path_b),
+        }
```
timon0305: self.allowed_resolved = {
    os.path.realpath(self.path_a),
    os.path.realpath(self.path_b),
}

The comparison set wraps each path in os.path.realpath(), but production resolve_workspace_path() calls expand_tilde_path() (verified at utils/path_helpers.py:14-25), which only does tilde expansion — it does not call realpath. On Linux this is a no-op because /tmp isn't a symlink, so the test passes. On macOS where /tmp → /private/tmp is a symlink, os.path.realpath(self.path_a) returns /private/tmp/cursor-ws-…/storage-a but resolve_workspace_path() returns /tmp/cursor-ws-…/storage-a — the resolved string is never in allowed_resolved and every reader iteration logs an error. Test will fail on macOS CI / dev machines.

Either drop the realpath calls:
self.allowed_resolved = {self.path_a, self.path_b}

or


self.tmp = tempfile.mkdtemp(prefix="cursor-ws-thread-test-",
                            dir=os.path.realpath(tempfile.gettempdir()))

utils/workspace_path.py
```diff
@@ -5,20 +5,27 @@
...
 import subprocess
+import threading
 
 from .path_helpers import expand_tilde_path
 
-# Module-level override set via the /api/set-workspace endpoint
+# Module-level override set via POST /api/set-workspace (or --base-dir).
+# All reads and writes are serialized by _workspace_path_lock so threaded
+# WSGI workers (gunicorn --threads, waitress, etc.) cannot observe torn
+# state between set_workspace_path_override and resolve_workspace_path.
```
timon0305: "Torn state" implies low-level word tearing of the reference, which CPython's GIL makes impossible for a single-pointer assignment. What the lock actually buys: (a) memory ordering / visibility across threads, and (b) atomicity of resolve_workspace_path's read-then-expand pair. Suggested rewording: "… so readers always observe the latest override set by another thread, and resolve_workspace_path's snapshot+expand sequence stays consistent."

tests/test_workspace_path_thread_safety.py
```diff
@@ -0,0 +1,118 @@
...
+        self._prior_workspace_env = os.environ.pop("WORKSPACE_PATH", None)
+        self.addCleanup(self._restore_workspace_env)
+        self.addCleanup(set_workspace_path_override, None)
+        set_workspace_path_override(None)
```
timon0305: The bare call on line 43 is redundant because (a) the cleanup from the previous test already cleared the override, and (b) line 56 (set_workspace_path_override(self.path_a)) immediately overwrites it before the workers start. Could drop line 44.

tests/test_workspace_path_thread_safety.py
```diff
@@ -0,0 +1,118 @@
...
+        self.fallback_resolved = resolve_workspace_path()
```
timon0305, line 44: Captured before any threads run. test_concurrent_clear_and_set_stays_consistent later allows resolved == self.fallback_resolved (:96-99), which works because WORKSPACE_PATH env was popped on :40. Fine as-is, but a one-line comment near :44 explaining why this snapshot is the legitimate "override cleared" result would make the assertion in the toggler test self-documenting. Optional.

timon0305 (approved):

tests/test_workspace_path_thread_safety.py
bradjin8: Used the raw paths the writer sets (same strings resolve_workspace_path() returns for these overrides)

utils/workspace_path.py
bradjin8: Updated the comment per the review. “Torn state” was misleading here — CPython’s GIL already makes a single pointer assignment atomic; the lock is really about cross-thread visibility and keeping resolve_workspace_path()’s read-then-expand_tilde_path() sequence consistent.

tests/test_workspace_path_thread_safety.py
bradjin8: Added a short comment above fallback_resolved noting that with WORKSPACE_PATH popped and no override, that snapshot is the expected path when the toggler test clears the override — so the resolved == self.fallback_resolved check in test_concurrent_clear_and_set_stays_consistent is easier to follow.

wpak-ai (approved):

PR #81, opened 2026-05-27
Add Hypothesis property tests for blob/bubble parsing (#71)

bradjin8:
## Summary

- Adds `tests/test_blob_parsing_fuzz.py` with Hypothesis property tests for the blob/bubble parsing pipeline (closes #71).
- Covers `Bubble.from_dict` (including adversarial non-dict inputs and invalid `bubble_id` types), workspace-tabs-style KV assembly (`json.loads` → `Bubble.from_dict`), CLI blob chain parsing (`classify_blob_data`, `_extract_blob_refs`, `traverse_blobs` on generated `store.db` fixtures), and text extraction (`extract_text_from_bubble`, `messages_to_bubbles`).
- Extracts shared `classify_blob_data()` in `utils/cli_chat_reader.py` so `traverse_blobs` and fuzz tests use the same JSON-vs-chain classification logic.
- Fixes `extract_text_from_bubble` to always return `str` when `text` is a non-string truthy value.
- Adds `hypothesis>=6.100,<7` to `[project.optional-dependencies].dev` in `pyproject.toml` and installs it in CI alongside pytest.
- Documents `pip install -e ".[dev]"` in the README and ignores `.hypothesis/` in `.gitignore`.

## Test plan

- [x] `python -m unittest tests.test_blob_parsing_fuzz -v`
- [x] `python -m unittest discover tests -v` (full suite, including new fuzz tests)
- [ ] CI green on the unittest matrix (Linux / Windows / macOS × Python 3.10–3.13)

## Notes

- Dev dependencies (`hypothesis`, `pytest`, `mypy`) are not added to `requirements.txt` / `requirements-lock.txt` (same pattern as pytest today).
- Eval context: addresses Test 31 / “Unverified Dark Path” coverage for parser invariants without crashing on schema drift inputs.

## Summary by CodeRabbit

* **Documentation**
  * Added development setup instructions to README with pip install guidance for contributors

* **Bug Fixes**
  * More robust text extraction with consistent string handling
  * Improved blob parsing and traversal for more reliable message/attachment handling

* **Tests**
  * Added comprehensive property/fuzz tests covering parsing, blob traversal, and text extraction

* **Chores**
  * CI test workflow now installs property-testing tooling
  * .gitignore updated to exclude local Hypothesis artifacts

[Review Change Stack](https://app.coderabbit.ai/change-stack/cppalliance/cppa-cursor-browser/pull/81?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -105,6 +105,7 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           python -m pip install -r requirements-lock.txt
+          # Dev-only (not in requirements-lock.txt): pytest + hypothesis for unittest/property tests.
           python -m pip install 'pytest>=8,<9' 'hypothesis>=6.100,<7'
 
       - name: Run unittest suite
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -98,14 +98,13 @@ jobs:
 
       - name: Install runtime + test dependencies
         # Install from the pinned lock file for deterministic dependency
-        # resolution (closes #47). pytest is added on top — it is not in
-        # requirements-lock.txt because it is a dev-only dep. pywebview is
+        # resolution (closes #47). pytest and hypothesis are added on top — not in
+        # requirements-lock.txt (dev-only). pywebview is
         # the desktop-launcher dep and pulls GTK / Qt system libraries on
         # Linux — intentionally excluded from the CI unittest matrix.
         run: |
           python -m pip install --upgrade pip
           python -m pip install -r requirements-lock.txt
-          # Dev-only (not in requirements-lock.txt): pytest + hypothesis for unittest/property tests.
           python -m pip install 'pytest>=8,<9' 'hypothesis>=6.100,<7'
 
       - name: Run unittest suite
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -105,7 +105,7 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           python -m pip install -r requirements-lock.txt
-          python -m pip install 'pytest>=8,<9'
+          python -m pip install 'pytest>=8,<9' 'hypothesis>=6.100,<7'
 
       - name: Run unittest suite
         run: python -m unittest discover tests -v
--- a/tests/test_blob_parsing_fuzz.py
+++ b/tests/test_blob_parsing_fuzz.py
@@ -102,6 +102,8 @@ _BUBBLE_LIKE = st.dictionaries(
     st.sampled_from(["text", "richText", "codeBlocks", "type", "metadata"]),
     st.one_of(
         st.text(max_size=300),
+        st.integers(),
+        st.booleans(),
         st.none(),
         st.lists(
             st.dictionaries(
--- a/tests/test_blob_parsing_fuzz.py
+++ b/tests/test_blob_parsing_fuzz.py
@@ -1,0 +1,190 @@
+"""Property-based fuzz tests for blob / bubble parsing (issue #71).
+
+Run:
+  python -m unittest tests.test_blob_parsing_fuzz -v
+  python -m pytest tests/test_blob_parsing_fuzz.py -v
+"""
+
+from __future__ import annotations
+
+import json
+import os
+import sys
+import unittest
+
+from hypothesis import given, settings
+from hypothesis import strategies as st
+
+REPO_ROOT = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+if REPO_ROOT not in sys.path:
+    sys.path.insert(0, REPO_ROOT)
+
+from models import Bubble, SchemaError
+from utils.cli_chat_reader import _extract_blob_refs, messages_to_bubbles
+from utils.text_extract import extract_text_from_bubble
+
+# Bounded strategies: fast enough for CI (<30s total with default example counts).
+_JSON_VALUES = st.one_of(
+    st.none(),
+    st.booleans(),
+    st.integers(),
+    st.floats(allow_nan=False, allow_infinity=False),
+    st.text(max_size=200),
+    st.lists(st.text(max_size=80), max_size=8),
+)
+
+_BUBBLE_RAW = st.dictionaries(
+    st.text(min_size=0, max_size=40),
+    _JSON_VALUES,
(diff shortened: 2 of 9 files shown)


tests/test_blob_parsing_fuzz.py
```diff
@@ -0,0 +1,331 @@
...
+        elif isinstance(value, str):
+            parsed = json.loads(value)
+        else:
+            parsed = value
+    except (json.JSONDecodeError, TypeError, ValueError, UnicodeDecodeError):
+        return None
+    try:
+        return Bubble.from_dict(parsed, bubble_id=bubble_id).raw  # type: ignore[arg-type]
+    except SchemaError:
+        return None
```
timon0305, line 186: _assemble_workspace_bubble doesn't truly mirror production.

The docstring promises "Keep in sync with the bubbleId load loop in services/workspace_tabs.py", but the test path diverges:

- Production (services/workspace_tabs.py:145) calls json.loads(row["value"]) unconditionally; whatever SQLite returns goes straight in.
- Test path branches: bytes/bytearray → json.loads(bytes(v).decode()), str → json.loads(v), else → parsed = value (skips json.loads entirely).

utils/cli_chat_reader.py
```diff
@@ -79,6 +79,23 @@ def _extract_blob_refs(data: bytes) -> list[str]:
...
+    Returns ``(message_dict, [])`` when *data* decodes to a dict with a
+    ``role`` field; otherwise ``(None, refs)`` where *refs* are SHA-256 hex
+    ids from :func:`_extract_blob_refs`.  Used by :func:`traverse_blobs` and
+    property tests — keep in sync when the load loop changes.
+    """
+    try:
+        msg = json.loads(data.decode("utf-8"))
+        if isinstance(msg, dict) and "role" in msg:
+            return msg, []
+    except (UnicodeDecodeError, json.JSONDecodeError, TypeError):
```
timon0305: classify_blob_data widens the except list unnecessarily

Original traverse_blobs caught (UnicodeDecodeError, json.JSONDecodeError). The extracted classify_blob_data adds TypeError:


except (UnicodeDecodeError, json.JSONDecodeError, TypeError):
But the only production caller (traverse_blobs:136-138) already guards with isinstance(data, bytes), so data.decode("utf-8") cannot TypeError. The fuzz test also feeds bytes (st.binary(...)). The widened catch is unexercised. Either drop it (parity with original behavior) or add a one-line comment noting it's defensive-only — otherwise it reads like a hidden semantic change in the refactor.

tests/test_blob_parsing_fuzz.py
```diff
@@ -0,0 +1,331 @@
...
+                st.text(max_size=20),
+                st.one_of(st.text(max_size=100), st.integers()),
+                max_size=5,
+            ),
+            max_size=4,
+        ),
+        st.dictionaries(st.text(max_size=20), _JSON_VALUES, max_size=5),
+    ),
+    max_size=6,
+)
```
timon0305, line 119: _BUBBLE_LIKE doesn't directly cover the bug text_extract.py fixes

The motivating fix is text = str(bubble["text"]) (coerce non-str). But _BUBBLE_LIKE's value one_of is text | None | list[dict] | dict[str, _JSON_VALUES] — no scalar (int/bool/float) directly under bubble["text"]. The fix is exercised indirectly via list/dict text inputs, but a literal {"text": 42} regression input would make the link between bug and test explicit. One-line tightening:


st.one_of(
    st.text(max_size=300),
    st.integers(),       # <-- exercises the str() coercion directly
    st.booleans(),
    st.none(),
    ...
)

tests/test_blob_parsing_fuzz.py
```diff
@@ -0,0 +1,331 @@
...
+REPO_ROOT = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+if REPO_ROOT not in sys.path:
+    sys.path.insert(0, REPO_ROOT)
+
+from models import Bubble, SchemaError
+from utils.cli_chat_reader import (
+    classify_blob_data,
+    messages_to_bubbles,
+    traverse_blobs,
+    _extract_blob_refs,  # internal helper; covered directly alongside classify_blob_data
```
timon0305: The import block grabs _extract_blob_refs directly (leading underscore = internal by convention). Justification comment is fine, but since the PR just extracted classify_blob_data as a peer public helper, promoting _extract_blob_refs → extract_blob_refs would let the test consume only the public surface. Trade-off either way; flagging because the next refactor may break this test for an unrelated reason.

tests/test_blob_parsing_fuzz.py
```diff
@@ -0,0 +1,331 @@
...
+            if bid not in blobs:
+                blobs[bid] = payloads[(i + 1) % len(payloads)]
+        with tempfile.TemporaryDirectory() as td:
+            db_path = os.path.join(td, "store.db")
+            _build_store_db_raw(db_path, meta, blobs)
+            try:
+                messages = traverse_blobs(db_path)
+            except Exception as exc:
+                self.fail(f"traverse_blobs raised {type(exc).__name__}: {exc}")
+            self.assertIsInstance(messages, list)
```
timon0305, line 296: test_traverse_blobs_never_raises skips the meta-parse error path

The meta blob is always built well-formed ({"latestRootBlobId": root_id, "createdAt": 1_700_000_000_000}). The CliSessionMeta.from_dict → SchemaError → return [] path in traverse_blobs:118-128 is never exercised. Quick add would be a separate @given(meta=st.dictionaries(...)) test that fuzzes the meta row too — small coverage gap on a real failure mode.

.github/workflows/tests.yml
```diff
@@ -105,7 +105,8 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           python -m pip install -r requirements-lock.txt
-          python -m pip install 'pytest>=8,<9'
+          # Dev-only (not in requirements-lock.txt): pytest + hypothesis for unittest/property tests.
```
timon0305: comment slightly misleading

# Dev-only (not in requirements-lock.txt): pytest + hypothesis for unittest/property tests.
python -m pip install 'pytest>=8,<9' 'hypothesis>=6.100,<7'

timon0305 (approved):

wpak-ai (approved):

PR #96, opened 2026-06-04
refactor(services): centralize workspace determination in resolve_workspace_context (#91)

bradjin8:
## Summary

- Adds `services/workspace_context.py` with a typed `WorkspaceContext` dataclass and `resolve_workspace_context()` orchestrator that runs the shared workspace-determination ceremony (entries, invalid IDs, project-name/path maps, composer→workspace map).
- Adds `enrich_workspace_context_from_global_db()` for optional global KV maps (`project_layouts_map`, `bubble_map`) after `open_global_db`.
- Refactors all four consumers to use the shared orchestrator instead of duplicating the 6-function setup:
  - `services/workspace_listing.py`
  - `services/workspace_tabs.py` (summaries, single-tab, full tabs)
  - `scripts/export.py`
  - `api/export_api.py` (minimal context: skips invalid/path maps; enriches bubbles only)
- Adds `tests/test_workspace_context.py` for orchestrator behavior and flag handling.

Closes #91.

## Test plan

- [x] `python -m pytest tests/test_workspace_context.py -q`
- [x] `python -m pytest -q` (full suite: 393 passed, 4 skipped)
- [ ] Manual smoke: `GET /api/workspaces` and `GET /api/workspaces/<id>/tabs` unchanged
- [ ] Manual smoke: `POST /api/export` and `python scripts/export.py` still assign chats to the expected workspace folders

## Summary by CodeRabbit

* **Refactor**
  * Centralized workspace discovery and optional global-database enrichment into a unified workspace context service for consistent resolution.

* **Refactor**
  * Export tooling and CLI now consume the simplified workspace context (including a minimal variant for HTTP export) instead of bespoke scanning logic.

* **Refactor**
  * Workspace listing and tab assembly now read mappings from the centralized context; message/context decoding and project layout extraction standardized.

* **Tests**
  * Added end-to-end tests for context resolution, caching behavior, pre-collected entries, invalid-workspace detection, and DB-driven enrichment.

### DIFF
--- a/api/export_api.py
+++ b/api/export_api.py
@@ -22,11 +22,9 @@ from utils.path_helpers import to_epoch_ms
 from utils.text_extract import extract_text_from_bubble, slug
 from utils.exclusion_rules import build_searchable_text, is_excluded_by_rules
 from utils.cursor_md_exporter import cursor_ide_chat_to_markdown
-from services.workspace_context import (
-    enrich_workspace_context_from_global_db,
-    resolve_workspace_context,
-)
+from services.workspace_context import resolve_workspace_context_minimal
 from services.workspace_db import (
+    load_bubble_map,
     load_code_block_diff_map,
     open_global_db,
 )
@@ -100,11 +98,7 @@ def export_chats():
                 last_export_ms = to_epoch_ms(ts_str)
 
         # ── Workspace scanning via service layer ──────────────────────────────
-        ctx = resolve_workspace_context(
-            workspace_path,
-            include_invalid_workspace_ids=False,
-            include_workspace_path_map=False,
-        )
+        ctx = resolve_workspace_context_minimal(workspace_path)
         workspace_entries = ctx.workspace_entries
         composer_id_to_ws = ctx.composer_id_to_workspace_id
 
@@ -126,10 +120,7 @@ def export_chats():
             if global_db is None:
                 return jsonify({"error": "Cursor global storage not found"}), 404
 
-            ctx = enrich_workspace_context_from_global_db(
-                ctx, global_db, populate_bubble_map=True,
-            )
-            bubble_map = ctx.bubble_map
+            bubble_map = load_bubble_map(global_db)
             code_block_diff_map = load_code_block_diff_map(global_db)
 
             try:
--- a/api/export_api.py
+++ b/api/export_api.py
@@ -22,17 +22,15 @@ from utils.path_helpers import to_epoch_ms
 from utils.text_extract import extract_text_from_bubble, slug
 from utils.exclusion_rules import build_searchable_text, is_excluded_by_rules
 from utils.cursor_md_exporter import cursor_ide_chat_to_markdown
+from services.workspace_context import (
+    enrich_workspace_context_from_global_db,
+    resolve_workspace_context,
+)
 from services.workspace_db import (
-    build_composer_id_to_workspace_id,
-    collect_workspace_entries,
-    load_bubble_map,
     load_code_block_diff_map,
     open_global_db,
 )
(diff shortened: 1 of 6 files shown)


services/workspace_context.py
```diff
@@ -0,0 +1,133 @@
...
+
+if TYPE_CHECKING:
+    import sqlite3
+
+
+@dataclass(frozen=True)
+class WorkspaceContext:
+    """Precomputed workspace-resolution maps for conversation assignment."""
+
+    workspace_path: str
```
timon0305: Nothing reads workspace_path — each caller already has it locally. Probably worth dropping.

services/workspace_context.py
```diff
@@ -0,0 +1,133 @@
...
+def resolve_workspace_context(
+    workspace_path: str,
+    *,
+    workspace_entries: list[dict] | None = None,
+    rules: list | None = None,
+    nocache: bool = False,
+    use_composer_cache: bool = False,
+    include_invalid_workspace_ids: bool = True,
+    include_workspace_path_map: bool = True,
+    global_db: sqlite3.Connection | None = None,
```
timon0305: No caller actually hits these — the two consumers that need global-db maps open the DB later and go through enrich_* instead. Two paths to the same fields will drift over time; might be cleaner to just drop the global_db params here.

services/workspace_context.py
```diff
@@ -0,0 +1,133 @@
...
+    populate_project_layouts: bool = False,
```
timon0305: This is a lot of toggles. nocache does nothing if use_composer_cache=False, and the include/use flags are easy to mix up. Two narrower entry points (one for listing, one minimal for export) might be friendlier than a single configurable one.

api/export_api.py
```diff
@@ -102,9 +100,13 @@ def export_chats():
                 last_export_ms = to_epoch_ms(ts_str)
 
         # ── Workspace scanning via service layer ──────────────────────────────
-        workspace_entries = collect_workspace_entries(workspace_path)
-        composer_id_to_ws = build_composer_id_to_workspace_id(workspace_path, workspace_entries)
-        project_name_map = create_project_name_to_workspace_id_map(workspace_entries)
+        ctx = resolve_workspace_context(
```
timon0305: Before this it was one call (load_bubble_map); now it's two. The listing/tabs paths win from the orchestrator but this one ends up longer. Not blocking, just noting.

tests/test_workspace_context.py
```diff
@@ -0,0 +1,102 @@
...
+        )
+        conn.commit()
+        try:
+            enriched = enrich_workspace_context_from_global_db(
+                ctx, conn, populate_bubble_map=True,
+            )
+        finally:
+            conn.close()
+        assert enriched.bubble_map.get("bid1") is not None
+        assert ctx.bubble_map == {}
```
timon0305, line 174: A few gaps: use_composer_cache=True only has the error case, populate_project_layouts=True isn't exercised, and there's no positive check that include_invalid_workspace_ids=True actually fills the set. Worth adding while the file is small.

timon0305 (commented):
Nice consolidation — four call sites going through one orchestrator reads cleanly. A few thoughts on the surface area inline.

tests/test_workspace_context.py
bradjin8: addressed (mapped to the refactored API: resolve_workspace_context_cached instead of use_composer_cache=True, full resolve instead of include_invalid_workspace_ids=True).

timon0305 (approved):

wpak-ai (approved):

PR #113, opened 2026-06-23
Speed up workspace listing and search; add FTS index and fix search d…

clean6378-max-it:
CLoses #95 

## Summary

Improves workspace list/navigation performance and makes search fast enough for daily use, while fixing incorrect conversation links from search results.

### Workspace listing & navigation (#95)
- Extract shared composer parsing/assignment into `workspace_composer_scan.py`
- Parallelize `build_composer_id_to_workspace_id` and bulk-load project layouts
- Align `conversationCount` with the same filters used by tab summaries
- Skip null/placeholder `composerData` rows without spurious warnings
- Restore `get_project_from_file_path` (fixes `NameError` that dropped ~66 conversations)

### Search performance
- **30-day default window** with UI checkbox and `all_history=1` API flag
- **Phase 1:** single-pass bubble `LIKE` scan (fixes scoped-query regression that made 30-day search slower than full history)
- **Phase 2:** local `search_index.sqlite` with FTS5, background rebuild on startup + 60s poll, fingerprint invalidation (same mtime scheme as listing cache)
- **Windows:** versioned index files + `search_index.active` pointer (avoids `PermissionError` when replacing an open DB)

### Search correctness
- Search `workspaceId` now uses `assign_composer_workspace` (same as sidebar), not roster map alone
- Workspace page honors `?tab=` deep links; fetches conversation directly instead of opening the first sidebar item; redirects to `global` on 404

## Test plan

- [ ] Restart `python app.py`; confirm log: `Search index rebuilt: N composers, M bubbles -> search_index.*.sqlite`
- [ ] Home page: projects load quickly; counts match sidebar conversation totals
- [ ] Search default (30 days): completes in a few seconds; checkbox searches all history
- [ ] Click a search result: opens the matching thread (not the first conversation in the project)
- [ ] `pytest tests/test_workspace_list_count_alignment.py tests/test_api_search.py tests/test_search_index.py tests/test_search_helpers.py tests/test_project_path_boundary.py`
- [ ] After Cursor adds a new chat, wait ~60s or restart; search finds new content

## Notes

- Index/cache lives under `~/.cache/cursor-chat-browser/`
- Bypass: `CURSOR_CHAT_BROWSER_NOCACHE=1` or `CURSOR_CHAT_BROWSER_NO_SEARCH_INDEX=1`
- Dev-only: `scripts/profile_search*.py`

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Added an “Include chats older than 30 days” search mode that expands results to all history.
  * Search API now supports a configurable history window and returns `allHistory` / `searchWindowDays` for accurate result labeling.

* **Performance/Background**
  * Search indexing can now start automatically during app initialization to keep search results current.

* **Bug Fixes**
  * Improved workspace tab navigation: missing tabs now redirect to the global workspace.

* **Tests**
  * Added/updated coverage for search windows, all-history API behavior, and end-to-end search index querying.

### DIFF
--- a/api/search.py
+++ b/api/search.py
@@ -1,6 +1,6 @@
 """
 API route for search — mirrors src/app/api/search/route.ts
-GET /api/search?q=...&type=all|chat|composer
+GET /api/search?q=...&type=all|chat|composer&all_history=1
 """
 
 import logging
@@ -12,7 +12,9 @@ from api.flask_config import json_response
 
 from models import ParseWarningCollector, SearchResult
 from services.search import (
+    DEFAULT_SEARCH_WINDOW_DAYS,
     rank_results,
+    resolve_search_since_ms,
     search_cli_sessions,
     search_global_storage,
     search_legacy_workspaces,
@@ -23,12 +25,26 @@ bp = Blueprint("search", __name__)
 _logger = logging.getLogger(__name__)
 
 
+def _parse_since_days_param(raw: str | None) -> int | None:
+    if raw is None or not str(raw).strip():
+        return None
+    try:
+        return int(raw)
+    except ValueError:
+        return None
+
+
 @bp.route("/api/search")
 def search() -> tuple[Response, int] | Response:
     try:
         query = request.args.get("q", "").strip()
         search_type = request.args.get("type", "all")
         rules = current_app.config.get("EXCLUSION_RULES") or []
+        all_history = request.args.get("all_history") in ("1", "true")
+        since_ms = resolve_search_since_ms(
+            all_history=all_history,
+            since_days=_parse_since_days_param(request.args.get("since_days")),
+        )
 
         if not query:
             return json_response({"error": "No search query provided"}, 400)
@@ -40,20 +56,46 @@ def search() -> tuple[Response, int] | Response:
         if search_type != "chat":
             results.extend(
                 search_global_storage(
-                    workspace_path, query, query_lower, rules, parse_warnings
+                    workspace_path,
+                    query,
+                    query_lower,
+                    rules,
+                    parse_warnings,
+                    since_ms=since_ms,
                 )
             )
         results.extend(
-            search_legacy_workspaces(workspace_path, query, query_lower, search_type, rules)
+            search_legacy_workspaces(
+                workspace_path,
+                query,
+                query_lower,
+                search_type,
+                rules,
+                since_ms=since_ms,
+            )
         )
         if search_type == "all":
             results.extend(
                 search_cli_sessions(
-                    get_cli_chats_path(), query, query_lower, rules, parse_warnings
+                    get_cli_chats_path(),
+                    query,
+                    query_lower,
+                    rules,
+                    parse_warnings,
+                    since_ms=since_ms,
                 )
             )
 
-        payload: dict[str, Any] = {"results": rank_results(results)}
+        payload: dict[str, Any] = {
+            "results": rank_results(results),
+            "allHistory": since_ms is None,
+            "searchWindowDays": (
+                None if since_ms is None else (
+                    _parse_since_days_param(request.args.get("since_days"))
+                    or DEFAULT_SEARCH_WINDOW_DAYS
+                )
+            ),
+        }
         return json_response(parse_warnings.attach_to(payload))
 
     except Exception:
--- a/api/search.py
+++ b/api/search.py
@@ -24,14 +24,19 @@ from utils.workspace_path import get_cli_chats_path, resolve_workspace_path
 bp = Blueprint("search", __name__)
 _logger = logging.getLogger(__name__)
 
+_MAX_SEARCH_SINCE_DAYS = 36_500  # ~100 years; avoids timedelta overflow on bad input
+
 
 def _parse_since_days_param(raw: str | None) -> int | None:
     if raw is None or not str(raw).strip():
         return None
     try:
-        return int(raw)
+        days = int(raw)
     except ValueError:
         return None
+    if days <= 0 or days > _MAX_SEARCH_SINCE_DAYS:
+        return None
+    return days
 
 
 @bp.route("/api/search")
--- a/services/search.py
+++ b/services/search.py
@@ -22,23 +22,34 @@ import logging
 import os
 import sqlite3
 from contextlib import closing
-from datetime import datetime
+from dataclasses import dataclass
+from datetime import datetime, timedelta, timezone
 from pathlib import Path
 from typing import Any
 
 __all__ = [
+    "DEFAULT_SEARCH_WINDOW_DAYS",
     "rank_results",
+    "resolve_search_since_ms",
     "search_cli_sessions",
     "search_global_storage",
     "search_legacy_workspaces",
 ]
 from models import Composer, ParseWarningCollector, SchemaError, SearchResult
+from services.workspace_composer_scan import assign_composer_workspace
+from services.workspace_context import (
+    WorkspaceContext,
+    enrich_workspace_context_from_global_db,
+    resolve_workspace_context_cached,
+)
 from services.workspace_db import (
+    COMPOSER_ROWS_WITH_HEADERS_SQL,
     build_composer_id_to_workspace_id_cached,
     collect_workspace_entries,
-    load_bubble_map,
     open_global_db,
+    safe_fetchall,
 )
+from services.workspace_resolver import infer_invalid_workspace_aliases
 from utils.cli_chat_reader import list_cli_projects, messages_to_bubbles, traverse_blobs
 from utils.exclusion_rules import build_searchable_text, is_excluded_by_rules
 from utils.path_helpers import (
@@ -46,12 +57,49 @@ from utils.path_helpers import (
     to_epoch_ms,
     warn_workspace_json_read,
 )
-from utils.text_extract import extract_text_from_bubble
 
 _logger = logging.getLogger(__name__)
 
 # Missing/unparseable timestamps sort last in rank_results() (treated as 0.0 s).
 _UNKNOWN_SEARCH_TIMESTAMP: int = 0
+
+DEFAULT_SEARCH_WINDOW_DAYS = 30
+
+# When a date window is active, chats with no parseable timestamp stay searchable.
+_INCLUDE_UNKNOWN_TIMESTAMPS_IN_WINDOW = True
+
+
+def resolve_search_since_ms(
+    *,
+    all_history: bool = False,
+    since_days: int | None = None,
+    now: datetime | None = None,
+) -> int | None:
+    """Return epoch-ms cutoff for search, or ``None`` to search all history."""
+    if all_history:
+        return None
+    days = since_days if since_days is not None else DEFAULT_SEARCH_WINDOW_DAYS
+    if days <= 0:
+        return None
+    ref = now or datetime.now(timezone.utc)
+    cutoff = ref - timedelta(days=days)
+    return int(cutoff.timestamp() * 1000)
+
+
+def _timestamp_in_search_window(timestamp_ms: int, since_ms: int | None) -> bool:
+    if since_ms is None:
+        return True
+    if timestamp_ms <= 0:
+        return _INCLUDE_UNKNOWN_TIMESTAMPS_IN_WINDOW
+    return timestamp_ms >= since_ms
+
+
+def _composer_dict_timestamp_ms(cd: dict[str, Any]) -> int:
+    return (
+        to_epoch_ms(cd.get("lastUpdatedAt"))
+        or to_epoch_ms(cd.get("createdAt"))
+        or 0
+    )
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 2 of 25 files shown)


services/search.py
```diff
@@ -154,6 +306,78 @@ def _build_ws_id_to_name(
...
+            composer,
+            project_layouts_map=self.ctx.project_layouts_map,
+            project_name_map=self.ctx.project_name_to_workspace_id,
+            workspace_path_map=self.ctx.workspace_path_to_id,
+            workspace_entries=self.ctx.workspace_entries,
+            bubble_map={},
+            composer_id_to_ws=self.ctx.composer_id_to_workspace_id,
+            invalid_workspace_ids=self.ctx.invalid_workspace_ids,
+            invalid_workspace_aliases=self.invalid_workspace_aliases,
+        )
```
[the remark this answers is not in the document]
clean6378-max-it: bubble_map={} in workspace assigner — Same pattern as workspace_tabs.py; loading full bubble maps would regress search performance.

services/workspace_tabs.py
bradjin8, line 1: ## services/workspace_tabs.py: line 578

assemble_single_tab still parses composerData with raw json.loads(row["value"]). A NULL value (e.g. composerData:empty-state-draft) raises TypeError, logs a decode warning, and returns HTTP 500. List/summary paths use parse_composer_data_row and skip these silently. Use the shared helper here for consistency with issue #95.

## services/workspace_tabs.py: line 623

Workspace ownership is checked with determine_project_for_conversation(..., bubble_map={}, ...) before scoped bubbles are loaded at line 637. Bubble-attached file paths cannot influence assignment on this path, so a conversation resolvable only via bubble metadata may 404 here even though list/summary/search agree on "global". Consider loading scoped bubbles before assignment, or document that this path intentionally matches the empty-bubble_map list/summary behavior.

## services/workspace_tabs.py: line 611
When invalid_workspace_ids is non-empty, every single-tab fetch runs safe_fetchall(global_db, COMPOSER_ROWS_WITH_HEADERS_SQL) — a full global composer scan : just to resolve aliases. On large installs this makes deep-link loads slow. Consider caching alias inference in the workspace context (same fingerprint as summary_cache).

services/workspace_tabs.py
```diff
@@ -814,34 +755,12 @@ def assemble_workspace_tabs(
...
-                )
-                parse_warnings.record_composer_skipped()
+            composer = parse_composer_data_row(
+                row["key"], row["value"], parse_warnings=parse_warnings,
+            )
+            if composer is None:
                 continue
+            composer_id = composer.composer_id
             try:
                 # Determine project
```
bradjin8: assemble_workspace_tabs inlines determine_project_for_conversation + alias logic instead of calling assign_composer_workspace. It also passes a loaded bubble_map, while list_workspace_tab_summaries (line 478) uses bubble_map={}. Full /tabs responses can therefore include conversations the summary sidebar omits (or vice versa) when assignment depends on bubble content. Either align assignment inputs or add a regression test documenting the intentional difference.

services/search.py
```diff
@@ -154,6 +339,78 @@ def _build_ws_id_to_name(
...
+    invalid_workspace_aliases: dict[str, str]
+
+    def workspace_for_composer(self, composer: Composer) -> str:
+        return assign_composer_workspace(
+            composer,
+            project_layouts_map=self.ctx.project_layouts_map,
+            project_name_map=self.ctx.project_name_to_workspace_id,
+            workspace_path_map=self.ctx.workspace_path_to_id,
+            workspace_entries=self.ctx.workspace_entries,
+            bubble_map={},
```
bradjin8, line 362: Workspace ownership is checked with determine_project_for_conversation(..., bubble_map={}, ...) before scoped bubbles are loaded at line 637. Bubble-attached file paths cannot influence assignment on this path, so a conversation resolvable only via bubble metadata may 404 here even though list/summary/search agree on "global". Consider loading scoped bubbles before assignment, or document that this path intentionally matches the empty-bubble_map list/summary behavior.

services/workspace_tabs.py
```diff
@@ -794,13 +741,7 @@ def assemble_workspace_tabs(
...
-        composer_rows = safe_fetchall(
-            global_db,
-            "SELECT key, value FROM cursorDiskKV WHERE key LIKE 'composerData:%'"
-            " AND value IS NOT NULL"
-            " AND value LIKE '%fullConversationHeadersOnly%'"
-            " AND value NOT LIKE '%fullConversationHeadersOnly\":[]%'",
-        )
+        composer_rows = safe_fetchall(global_db, COMPOSER_ROWS_WITH_HEADERS_SQL)
 
         invalid_workspace_aliases = infer_invalid_workspace_aliases(
```
bradjin8: infer_invalid_workspace_aliases(...) is called unconditionally. The summary path guards with if invalid_workspace_ids: (line 458). When the set is empty this loop is cheap (immediate continue), but on large installs it still walks every composer row on every full /tabs request. Mirror the summary guard to skip the scan entirely.

services/search.py
```diff
@@ -176,104 +435,169 @@ def search_global_storage(
...
+                        )
+                    if not has_match:
                         continue
-                    text = extract_text_from_bubble(bubble)
-                    if text:
-                        bubble_texts.append(text)
-                    bubble_meta.append(_json_dump_safe(bubble.raw))
 
-                exclusion_text = _build_exclusion_searchable(
+                all_bubble_texts = query_all_composer_bubble_texts(composer_id)
```
bradjin8: In the indexed search path, query_all_composer_bubble_texts(composer_id) opens a read-only index connection per candidate hit. With many matches this becomes N+1 I/O. Hoist to one query (e.g. batch by composer_id IN (...)) or reuse bubble texts already returned from query_composer_bubble_hits when they cover the full exclusion check.

services/search_index.py
```diff
@@ -0,0 +1,536 @@
...
+                                composer_id,
+                                title,
+                                to_epoch_ms(cd.get("createdAt")) or 0,
+                                _composer_dict_timestamp_ms(cd),
+                                raw_text,
+                            ),
+                        )
+                        composer_count += 1
+
+                    bubble_rows = src_conn.execute(
```
bradjin8, line 297: Bubble indexing iterates all bubbleId:% rows without checking that the parent composer exists in the composers table. Orphan FTS rows add rebuild cost and can produce bubble hits that are dropped later when composer_row is missing from search_pool (line 525). Filter bubbles to indexed composer IDs during build.

services/workspace_listing.py
```diff
@@ -161,54 +135,35 @@ def _build_workspace_projects_uncached(
...
-                            project_name_map, workspace_path_map,
-                            workspace_entries, bubble_map, composer_id_to_ws, invalid_workspace_ids,
+                        assigned = assign_composer_workspace(
+                            composer,
+                            project_layouts_map=project_layouts_map,
+                            project_name_map=project_name_map,
+                            workspace_path_map=workspace_path_map,
+                            workspace_entries=workspace_entries,
+                            bubble_map=bubble_map,
+                            composer_id_to_ws=composer_id_to_ws,
```
bradjin8, line 156: bubble_map={} is passed to assign_composer_workspace for performance. Conversations resolvable only via bubble file paths will land in "global" and may depress per-project counts. This is consistent with summary, but worth a code comment explaining the deliberate tradeoff so future readers don't "fix" it by loading the global bubble index on the list path.

services/workspace_resolver.py
```diff
@@ -63,100 +62,121 @@ def lookup_workspace_display_name(workspace_path: str, workspace_id: str) -> str
...
+        return matching
+    target_folder = ""
+    wj_path = os.path.join(workspace_path, workspace_id, "workspace.json")
+    try:
+        wd = read_json_file(wj_path)
+        folders = get_workspace_folder_paths(wd)
+        first_folder = folders[0] if folders else None
+        if first_folder:
+            target_folder = normalize_file_path(first_folder)
+    except Exception as e:
```
bradjin8: Broad except Exception around workspace.json reads (also line 123, and line 436 in infer_invalid_workspace_aliases) swallows unexpected defects. Catch (OSError, ValueError, TypeError, KeyError) (or the specific errors from read_json_file) so programming errors still surface during development.

services/workspace_tabs.py
bradjin8: Replace the duplicated assignment block with assign_composer_workspace(...) (as list/summary already do) to prevent future drift in alias handling and "global" fallback logic.

tests/test_workspace_list_count_alignment.py
```diff
@@ -0,0 +1,125 @@
...
+    card = next(p for p in projects if p["id"] == HAPPY_WORKSPACE_ID)
+
+    payload, status = list_workspace_tab_summaries(
+        HAPPY_WORKSPACE_ID, ws_root, rules=[], nocache=True,
+    )
+    assert status == 200
+    assert card["conversationCount"] == len(payload["tabs"])
+
+
+def test_null_composer_placeholder_skipped_silently():
```
bradjin8, line 90: Tests cover parse_composer_data_row for null placeholders but not assemble_single_tab or the full /tabs endpoint with a NULL composerData:empty-state-draft row. Add an integration test ensuring those paths return 404/empty without warnings, not HTTP 500.

tests/test_workspace_list_count_alignment.py
bradjin8, line 1: No test verifies summary vs full /tabs assignment parity when workspace resolution depends on bubble-attached file paths. Even if the behavior is intentionally different, a test would lock the contract and prevent silent regressions.
New test suggested.

services/search.py
```diff
@@ -22,37 +22,88 @@
...
-from utils.text_extract import extract_text_from_bubble
 
 _logger = logging.getLogger(__name__)
 
 # Missing/unparseable timestamps sort last in rank_results() (treated as 0.0 s).
 _UNKNOWN_SEARCH_TIMESTAMP: int = 0
 
+DEFAULT_SEARCH_WINDOW_DAYS = 30
+
+# When a date window is active, chats with no parseable timestamp stay searchable.
```
bradjin8, line 69: _INCLUDE_UNKNOWN_TIMESTAMPS_IN_WINDOW = True means composers with updated_ms <= 0 always pass the 30-day window. That matches index SQL (updated_ms <= 0) but may surprise users who expect a strict cutoff. Consider documenting in the search UI tooltip or API docs.

api/search.py
```diff
@@ -40,20 +61,46 @@ def search() -> tuple[Response, int] | Response:
...
+            "searchWindowDays": (
+                None if since_ms is None else (
+                    _parse_since_days_param(request.args.get("since_days"))
+                    or DEFAULT_SEARCH_WINDOW_DAYS
+                )
+            ),
+        }
         return json_response(parse_warnings.attach_to(payload))
 
     except Exception:
```
bradjin8, line 106: The outer except Exception returns a generic 500 for all failures. Consider distinguishing bad input (400) from infrastructure errors so clients can retry appropriately.

bradjin8 (changes requested):

services/workspace_tabs.py
clean6378-max-it: Cache alias inference in workspace context: Too Larger change; full composer scan in assemble_single_tab unchanged
Need to create another seperate PR for this. Out of scope

services/search.py
clean6378-max-it: Doc-only in API layer for now

services/workspace_tabs.py
clean6378-max-it: Good catch — agreed. assemble_single_tab now uses parse_composer_data_row like list/summary/full-tabs. A NULL composerData:empty-state-draft row returns 404 instead of logging a decode error and returning 500. Added test_single_tab_null_composer_placeholder_returns_404 in test_workspace_list_count_alignment.py.

clean6378-max-it: Valid concern for deep links. For single-tab we now load scoped bubbles before workspace assignment and pass that map into assign_composer_workspace. That’s cheap (one composer) and avoids a 404 when resolution depends on bubble-attached paths. List/summary/search still use an empty bubble map on purpose for performance - see the comment on the listing/search assigner paths.

services/workspace_tabs.py
clean6378-max-it: Fixed. Full /tabs now calls assign_composer_workspace with bubble_map={} for assignment, same as summary. The global bubble_map is still loaded for tab body assembly (_assemble_tab_from_composer_data) only, so we don’t regress list/summary parity on which conversations belong to which workspace.

services/search.py
clean6378-max-it: Same fix as above: assignment uses bubble_map={}; loaded map is only for rendering bubbles. Added test_summary_and_full_tabs_share_assignment (with a minimal bubble row so full tabs can assemble) to lock the contract.

services/workspace_tabs.py
clean6378-max-it: Fixed — wrapped in if invalid_workspace_ids: like the summary path, so we skip the alias walk entirely when there are no invalid workspaces.

services/search.py
clean6378-max-it: Intentional, same tradeoff as list/summary: we don’t load the global bubble index on the search hot path. Search workspace IDs go through assign_composer_workspace (not roster map alone). Added a short comment on workspace_for_composer so it doesn’t get “fixed” back to a full bubble scan later.

services/workspace_listing.py
clean6378-max-it: Added a comment at the assign_composer_workspace call: empty bubble map matches tab summary assignment (#95 perf tradeoff). Conversations resolvable only via bubble file paths may count as global — that’s deliberate and consistent across list/summary/search.

services/search_index.py
clean6378-max-it: Fixed. Index build collects indexed_composer_ids while inserting composers and skips bubble rows whose composer_id isn’t in that set.

services/workspace_resolver.py
clean6378-max-it: Narrowed to (OSError, json.JSONDecodeError, KeyError, TypeError, ValueError) in matching_workspace_ids_for_folder, and alias inference’s composer JSON decode to (json.JSONDecodeError, TypeError, ValueError) so unexpected bugs aren’t swallowed.

api/search.py
clean6378-max-it: Fair point. Left as-is for this PR to avoid widening error-handling scope
Will need to create a seperate PR

services/search.py
clean6378-max-it: Documented in resolve_search_since_ms: composers with no parseable timestamp stay searchable when a window is active (matches index SQL updated_ms <= 0). I didn’t add a UI tooltip in this PR 

Need a seperate PR

tests/test_workspace_list_count_alignment.py
clean6378-max-it: Added test_single_tab_null_composer_placeholder_returns_404. Full /tabs already goes through parse_composer_data_row from the earlier #95 migration.

tests/test_workspace_list_count_alignment.py
clean6378-max-it: Added test_summary_and_full_tabs_share_assignment with a seeded bubble so full tabs can render; asserts the same composer IDs on summary and full /tabs after the assign_composer_workspace + bubble_map={} alignment.

services/workspace_tabs.py
clean6378-max-it: Addressed in the latest push. assemble_workspace_tabs no longer inlines determine_project_for_conversation + alias remap; it calls assign_composer_workspace(...) like list/summary.

bradjin8 (approved):

wpak-ai (approved):

PR #112, opened 2026-06-22
Extract shared export engine; wire CLI to summary cache and drop mypy…

clean6378-max-it:
Closes #106

## Summary

Finishes the export-script consolidation started in #42 / PR #61 by extracting shared orchestration into `services/export_engine.py`, reducing `scripts/export.py` to a thin CLI wrapper, and routing the CLI through the same cached workspace-resolution path as the web layer (summary cache + mtime fingerprinting). Removes the `scripts.export` mypy strict override.

Continues cppa-cursor-browser #42. Companion PR (Monday #4): `os.environ` workspace-path bypass fix — land after this so explicit-override plumbing slots into the shared engine.

## Changes

- **New `services/export_engine.py`** — shared orchestration:
  - `prepare_workspace_orchestration()` — workspace scan, mtime fingerprint, `resolve_workspace_context_cached`
  - `load_global_db_export_data()` — global DB maps, composer rows, invalid-workspace aliases
  - `collect_export_entries()` — IDE + CLI conversation collection, exclusion filtering, markdown generation
- **`scripts/export.py`** — reduced to arg parsing, manifest/state I/O, zip/file output; delegates orchestration to `collect_export_entries()`
- **`services/workspace_listing.py`** — uses `prepare_workspace_orchestration()` instead of duplicating fingerprint + context setup
- **`pyproject.toml`** — removed `[[tool.mypy.overrides]]` `ignore_errors = true` for `scripts.export`

## Out of scope

- `os.environ["WORKSPACE_PATH"]` mutation for `--base-dir` (Monday PR 2, issue #4)
- CLI export output format changes
- Structured error codes / business-logic extraction / dual-model bridge

## Acceptance criteria

- [x] `scripts/export.py` calls shared export orchestration instead of reimplementing workspace scanning, DB reading, and composer processing
- [x] Summary cache and mtime fingerprinting available to CLI path
- [x] `scripts.export` mypy override removed
- [x] `scripts/export.py` passes `mypy --strict`
- [x] Existing CLI export tests pass
- [x] No behavioral change to CLI export output
- [ ] PR approved by at least 1 reviewer

## Test plan

- [x] `mypy --strict scripts/export.py services/export_engine.py services/workspace_listing.py`
- [x] `pytest tests/test_cli_export_e2e.py tests/test_export_exclusion_filtering.py tests/test_cli_args.py tests/test_api_export.py -q`
- [x] `pytest tests/test_api_workspaces.py tests/test_workspace_listing_*.py -q`
[...]

PR #114, opened 2026-06-23
Fix/export workspace path override

clean6378-max-it:
Closes #107 

## Summary

Replaces the CLI `--base-dir` workspace-path bypass that mutated `os.environ["WORKSPACE_PATH"]` with an explicit `resolve_workspace_path(override=...)` parameter. Stacks on Monday PR 1 (`refactor/export-engine-consolidation`); the core override plumbing landed there during review — this PR adds dedicated regression tests and a doc comment fix.

Sprint item Chen § #4 (3 pt, Medium).

## Changes

- **`utils/workspace_path.py`** — `resolve_workspace_path(*, override: str | None = None)`; call-time override takes precedence over module override, env, and default; does not mutate `WORKSPACE_PATH`
- **`scripts/export.py`** — `resolve_workspace_path(override=opts.get("base_dir"))`; `os.environ["WORKSPACE_PATH"]` mutation removed
- **`tests/test_export_base_dir_override.py`** (new) — asserts `--base-dir` is passed as `override=` and leaves `WORKSPACE_PATH` unchanged
- **`tests/test_workspace_path_thread_safety.py`** — explicit-override precedence test (from PR 1 stack)

## Out of scope

- Export-engine consolidation (Monday PR 1 — #3)
- Web server `WORKSPACE_PATH` lock semantics

## Acceptance criteria

- [x] `scripts/export.py` passes `--base-dir` explicitly to `resolve_workspace_path()` instead of mutating `os.environ`
- [x] `os.environ["WORKSPACE_PATH"]` mutation line removed
- [x] `resolve_workspace_path()` accepts optional `override` parameter
- [x] Existing export / workspace_path / thread-safety tests pass
- [x] Thread-safety tests unaffected
- [ ] PR approved by at least 1 reviewer

## Test plan

- [x] `mypy --strict utils/workspace_path.py scripts/export.py`
- [x] `pytest tests/test_export_base_dir_override.py tests/test_workspace_path_thread_safety.py tests/test_cli_export_e2e.py -q`
- [x] `pytest -k "workspace_path or export or thread" -q`
- [x] `python scripts/export.py --base-dir <workspaceStorage> --help`

## Merge note

**Base branch:** `refactor/export-engine-consolidation` (PR 1 must merge first, or merge both in order).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved export robustness by skipping invalid composer rows and treating corrupt stored values more safely.
  * Fixed CLI session exports to derive session creation timestamps from provided metadata when available.

* **Improvements**
  * Improved workspace project listing performance by adding cache-hit behavior when caching is enabled.
  * Enhanced `--base-dir` handling to correctly override workspace path resolution without permanently changing `WORKSPACE_PATH`.

* **Tests**
  * Added regression coverage for `--base-dir` override behavior and environment variable preservation.

### DIFF
--- a/api/export_api.py
+++ b/api/export_api.py
@@ -1,141 +1,142 @@
-"""
-API route for export — produces per-chat Markdown in a zip download.
-POST /api/export  { since: "all"|"last", zip: true }
-GET  /api/export/state — returns last export time
-"""
-
-from __future__ import annotations
-
-import io
-import json
-import logging
-import os
-import zipfile
-from datetime import datetime
-from pathlib import Path
-from typing import Any, Literal
-
-from flask import Blueprint, Response, request
-
-from api.flask_config import exclusion_rules, json_response
-from services.export_engine import collect_export_entries, read_last_export_ms
-from services.workspace_db import global_storage_db_path
-from utils.workspace_path import resolve_workspace_path
-
-bp = Blueprint("export_api", __name__)
-_logger = logging.getLogger(__name__)
-
-
-def _get_state_dir() -> str:
-    return os.path.join(str(Path.home()), ".cursor-chat-browser")
-
-
-def _get_export_state() -> dict[str, Any]:
-    """Read the export state file."""
-    state_path = os.path.join(_get_state_dir(), "export_state.json")
-    if os.path.isfile(state_path):
-        try:
-            with open(state_path, "r", encoding="utf-8") as f:
-                parsed = json.load(f)
-            if isinstance(parsed, dict):
-                return parsed
-            _logger.warning(
-                "Export state in %s is not a JSON object (got %s); ignoring",
-                state_path,
-                type(parsed).__name__,
-            )
-        except (json.JSONDecodeError, ValueError, OSError) as e:
-            _logger.warning(
-                "Could not read export state from %s: %s",
-                state_path,
-                e,
-            )
-    return {}
-
-
-def _save_export_state(count: int) -> None:
-    """Save export state after an export."""
-    state_dir = _get_state_dir()
-    os.makedirs(state_dir, exist_ok=True)
-    state = {
-        "lastExportTime": datetime.now().isoformat(),
-        "exportedCount": count,
-    }
-    state_path = os.path.join(state_dir, "export_state.json")
-    with open(state_path, "w", encoding="utf-8") as f:
-        json.dump(state, f, indent=2)
-
-
-@bp.route("/api/export/state")
-def get_export_state() -> Response:
-    """Return the last export timestamp."""
-    state = _get_export_state()
-    return json_response(state)
-
-
-@bp.route("/api/export", methods=["POST"])
-def export_chats() -> tuple[Response, int] | Response:
-    """Export chats as a zip archive.
-
-    Exclusion rules (``EXCLUSION_RULES`` app config key) are evaluated against
-    each chat's project name, title, and model.  Rules are loaded once at
-    application startup; an app restart is required to pick up changes to the
-    exclusion rules file.
-    """
-    try:
-        body = request.get_json(silent=True)
-        if not isinstance(body, dict):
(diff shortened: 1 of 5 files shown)


api/export_api.py
timon0305, line 1: This PR contains the entire #112 , and the new regression tests are exactly what was missing. The main issue is that the PR is opened against master and re-includes all of #112 's engine extraction.

tests/test_export_base_dir_override.py
timon0305, line 1: Nit: The 4-deep nested with blocks could use parenthesized context managers for readability.

scripts/export.py
```diff
@@ -172,318 +168,57 @@ def parse_args():
...
     out_dir = os.path.abspath(opts["out_dir"])
     use_zip = opts["zip"]
-    exclusion_rules = load_rules(resolve_exclusion_rules_path(opts.get("exclusion_rules_path")))
-    if opts.get("base_dir"):
-        os.environ["WORKSPACE_PATH"] = opts["base_dir"]
-    workspace_path = resolve_workspace_path()
+    exclusion_rules = load_rules(
+        resolve_exclusion_rules_path(opts.get("exclusion_rules_path")),
+    )
+    workspace_path = resolve_workspace_path(override=opts.get("base_dir"))
```
timon0305, line 179: This is the actual fix and it's correct. --base-dir no longer mutates the global env.

services/export_engine.py
```diff
@@ -0,0 +1,507 @@
...
+from utils.path_helpers import to_epoch_ms
+from utils.text_extract import extract_text_from_bubble, slug
+from utils.workspace_path import get_cli_chats_path
+
+_logger = logging.getLogger(__name__)
+
+SinceMode = Literal["all", "last"]
+
+
+class ExportEntry(TypedDict):
```
timon0305: Carried over from #112 : duplicates models.ExportEntry. Fix it in the PR that owns export_engine.py so it's not addressed twice. Import from models instead.

services/export_engine.py
```diff
@@ -0,0 +1,507 @@
...
+                composer_id,
+                type(cd).__name__,
+            )
+            continue
+
+        headers = cd.get("fullConversationHeadersOnly") or []
+        if not isinstance(headers, list) or not headers:
+            continue
+
+        updated_at = to_epoch_ms(cd.get("lastUpdatedAt")) or to_epoch_ms(
```
timon0305: Also carried from #112 : This now really falls back to createdAt, which changes --since last filtering and timestamps. Needs a test + a wording fix on "no behavior change." Same note as #112

timon0305 (changes requested):

api/export_api.py
clean6378-max-it: Keeping #114 as a thin follow-up stacked on this PR. The resolve_workspace_path(override=…) plumbing had to land here so --base-dir works through the shared engine; #114 adds the dedicated regression tests and doc clarification for sprint item #4 without mixing more scope into the consolidation PR. Will merge #112 first, then #114.

clean6378-max-it: Rebased was needed

services/export_engine.py
clean6378-max-it: Agreed — addressed in #112 (CollectedExportEntry in models/export.py). This branch stacks on #112; those engine changes land when #112 merges. Not re-fixing here to avoid duplicate commits across PRs.

services/export_engine.py
clean6378-max-it: Same — test + wording update are in #112 (test_created_at_fallback_when_last_updated_missing + PR description). Intentional behavior change; documented there.

scripts/export.py
clean6378-max-it: Thanks — that's the core of this PR. #114 adds the regression tests that lock it in

timon0305 (approved):

wpak-ai (approved):

PR #120, opened 2026-06-24
test: add summary-cache pytest-benchmark suite with CI regression gate

clean6378-max-it:
Closes #115 

## Summary

Adds a `pytest-benchmark` suite for `services/summary_cache.py` with synthetic fixtures and a CI regression gate. No production code changes.

- New `tests/benchmarks/test_summary_cache_bench.py` measures cache hit, cache miss, fingerprint cost (10/50/200 workspace entries), and round-trip write+read latency
- New `tests/benchmarks/conftest.py` builds synthetic workspace trees and isolates cache files to temp dirs (no real Cursor storage)
- `benchmarks/baselines.json` records initial gated means (1.5× slack from local reference run; refresh from ubuntu-latest CI artifact after first green job)
- `scripts/check_benchmark_regression.py` fails CI when any benchmark mean exceeds baseline by >20%
- `pyproject.toml`: add `pytest-benchmark` dev dep; default `--benchmark-skip` so normal `pytest` stays fast
- `.github/workflows/tests.yml`: new `benchmarks` job on ubuntu-latest

Sprint item **#7** (June Week 4, Wednesday). Sequences before Thursday **#8** (unified perf benchmark suite).

## Test plan

- [ ] `pytest tests/benchmarks/test_summary_cache_bench.py --benchmark-only -o addopts=`
- [ ] `python scripts/check_benchmark_regression.py benchmark-results.json benchmarks/baselines.json`
- [ ] `pytest tests/benchmarks/test_summary_cache_bench.py -o addopts=` (benchmarks skipped via `--benchmark-skip`)
- [ ] `python -m unittest discover tests -q`
- [ ] Confirm `benchmarks` CI job is green on ubuntu-latest

## Summary by CodeRabbit

* **New Features**
  * Added CI performance benchmarking for summary-cache (cache hit/miss, workspace fingerprinting, and cache round-trips).
* **Bug Fixes**
  * Updated test execution to avoid inheriting unrelated pytest options.
* **Chores**
  * Introduced a benchmark regression gate with updated baselines and CI artifact handling; refreshed benchmark baseline data and ignored generated benchmark outputs.
  * Enabled `pytest-benchmark` in dev tooling and configured pytest benchmark defaults.
* **Tests**
  * Added new benchmark test coverage and a dedicated unit test suite for the regression gate logic.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -115,7 +115,7 @@ jobs:
         # exercise Flask routes via app.test_client(). Only listed files — not
         # `pytest tests/` — to avoid re-collecting unittest.TestCase classes above.
         # -o addopts= avoids inheriting benchmark-only options from pyproject.toml.
-        run: python -m pytest tests/test_api_search.py tests/test_api_workspaces.py tests/test_api_export.py tests/test_pdf_export.py tests/test_search_helpers.py -v --tb=short -o addopts=
+        run: python -m pytest tests/test_api_search.py tests/test_api_workspaces.py tests/test_api_export.py tests/test_pdf_export.py tests/test_search_helpers.py tests/test_check_benchmark_regression.py -v --tb=short -o addopts=
 
       # ── PyInstaller desktop build (Windows only, once per workflow) ────────
       # Closes #44. Builds the onedir bundle and smoke-tests --help so the
@@ -240,6 +240,7 @@ jobs:
         run: >
           python -m pytest tests/benchmarks/
           --benchmark-only
+          --benchmark-min-rounds=5
           --benchmark-json=benchmark-results.json
           --benchmark-columns=min,max,mean,stddev,rounds
           -o addopts=
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -105,7 +105,7 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           python -m pip install -r requirements-lock.txt
-          python -m pip install 'pytest>=8,<9' 'hypothesis>=6.100,<7'
+          python -m pip install 'pytest>=8,<9' 'pytest-benchmark==4.0.0' 'hypothesis>=6.100,<7'
 
       - name: Run unittest suite
         run: python -m unittest discover tests -v
@@ -218,6 +218,7 @@ jobs:
   # ── Performance benchmarks: summary cache (issue #7) ───────────────────────
   benchmarks:
     name: Performance benchmarks (gated)
+    needs: [unittest]
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
@@ -233,7 +234,7 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           python -m pip install -r requirements-lock.txt
-          python -m pip install 'pytest>=8,<9' 'pytest-benchmark>=4,<5'
+          python -m pip install 'pytest>=8,<9' 'pytest-benchmark==4.0.0'
 
       - name: Run summary-cache benchmarks
         run: >
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -215,7 +215,7 @@ jobs:
             --redact \
             --exit-code 1
 
-  # ── Performance benchmarks: summary cache (issue #7) ───────────────────────
+  # ── Performance benchmarks: summary cache (issue #115) ─────────────────────
   benchmarks:
     name: Performance benchmarks (gated)
     needs: [unittest]
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -213,3 +213,40 @@ jobs:
             --verbose \
             --redact \
             --exit-code 1
+
+  # ── Performance benchmarks: summary cache (issue #7) ───────────────────────
+  benchmarks:
+    name: Performance benchmarks (gated)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+        with:
+          persist-credentials: false
+
+      - name: Set up Python
+        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405  # v6.2.0
+        with:
+          python-version: "3.12"
+
+      - name: Install runtime + benchmark dependencies
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install -r requirements-lock.txt
+          python -m pip install 'pytest>=8,<9' 'pytest-benchmark>=4,<5'
+
+      - name: Run summary-cache benchmarks
+        run: >
+          python -m pytest tests/benchmarks/
+          --benchmark-only
+          --benchmark-json=benchmark-results.json
+          --benchmark-columns=min,max,mean,stddev,rounds
+          -o addopts=
+
+      - name: Regression gate
+        run: python scripts/check_benchmark_regression.py benchmark-results.json benchmarks/baselines.json
+
+      - uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02  # v4
+        if: always()
+        with:
+          name: benchmark-results
+          path: benchmark-results.json
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -114,7 +114,8 @@ jobs:
         # Pytest fixtures (tests/conftest.py) build a temp workspaceStorage and
         # exercise Flask routes via app.test_client(). Only listed files — not
         # `pytest tests/` — to avoid re-collecting unittest.TestCase classes above.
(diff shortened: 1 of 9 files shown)


.github/workflows/tests.yml
```diff
@@ -114,7 +114,8 @@ jobs:
         # Pytest fixtures (tests/conftest.py) build a temp workspaceStorage and
         # exercise Flask routes via app.test_client(). Only listed files — not
         # `pytest tests/` — to avoid re-collecting unittest.TestCase classes above.
-        run: python -m pytest tests/test_api_search.py tests/test_api_workspaces.py tests/test_api_export.py tests/test_pdf_export.py tests/test_search_helpers.py -v --tb=short
+        # -o addopts= avoids inheriting benchmark-only options from pyproject.toml.
+        run: python -m pytest tests/test_api_search.py tests/test_api_workspaces.py tests/test_api_export.py tests/test_pdf_export.py tests/test_search_helpers.py -v --tb=short -o addopts=
```
bradjin8: tests/test_check_benchmark_regression.py (15 tests for the regression gate) is not run in CI. unittest discover won't collect pytest functions, and the integration pytest step lists only API/search files. Add this file to the pytest step or run pytest tests/test_check_benchmark_regression.py -o addopts= so gate logic is covered on every matrix run.

requirements-lock.txt
```diff
@@ -6,7 +6,7 @@
 # Lock is generated on Linux (CI / update-lock.yml). Windows-only transitives (e.g.
 # colorama via click) are omitted — pip still installs them on Windows when needed.
 blinker==1.9.0            # via flask
-click==8.4.1              # via flask
+click==8.4.2              # via flask
```
bradjin8, line 9: Unrelated click==8.4.1 → 8.4.2 bump is out of scope for a benchmarks-only PR. Revert unless there's a documented dependency reason; otherwise it adds review nois

pyproject.toml
```diff
@@ -31,10 +31,19 @@ desktop = ["pywebview>=5.0,<6"]
...
 dev = [
     "pytest>=8,<9",
+    "pytest-benchmark>=4,<5",
     "mypy>=1.10,<2",
     "hypothesis>=6.100,<7",
 ]
 
+[tool.pytest.ini_options]
+pythonpath = ["."]
+addopts = "--benchmark-disable"
```
bradjin8: PR description says --benchmark-skip but config uses --benchmark-disable. With disable, benchmark tests still run once on a bare pytest (functional smoke, ~0.4s). Either align the docs or switch to --benchmark-skip if the goal is zero benchmark execution during normal pytest.

pyproject.toml
```diff
@@ -31,10 +31,19 @@ desktop = ["pywebview>=5.0,<6"]
...
+testpaths = ["tests"]
```
bradjin8, line 43: New [tool.pytest.ini_options] with addopts = "--benchmark-disable" requires pytest-benchmark installed for any default pytest invocation. CI handles this; document in PR/issue that devs need pip install -e ".[dev]" (or pytest-benchmark) before running pytest without -o addopts=.

tests/benchmarks/test_summary_cache_bench.py
```diff
@@ -0,0 +1,74 @@
...
+@pytest.mark.benchmark(group="summary-cache")
+def test_summary_cache_miss(
+    benchmark,
+    summary_cache_dir: Path,
+    workspace_fingerprint: dict[str, Any],
+    stale_fingerprint: dict[str, Any],
+    sample_projects: list[dict[str, Any]],
+) -> None:
+    set_cached_projects(workspace_fingerprint, sample_projects, [])
+    benchmark(get_cached_projects, stale_fingerprint)
```
bradjin8, line 41: test_summary_cache_hit and test_summary_cache_miss measure nearly identical paths (~63µs on CI); both baselines are 6.3e-05. Consider merging into one "cache lookup" benchmark with hit/miss as subcases, or document that miss means fingerprint-mismatch only (not rebuild cost).

tests/benchmarks/conftest.py
```diff
@@ -0,0 +1,88 @@
...
+    """
+    cache_dir = tmp_path / "cache"
+    cache_dir.mkdir()
+    monkeypatch.setattr(summary_cache, "CACHE_DIR", cache_dir)
+    monkeypatch.setattr(summary_cache, "PROJECTS_CACHE_FILE", cache_dir / "projects.json")
+    monkeypatch.setattr(
+        summary_cache,
+        "COMPOSER_MAP_CACHE_FILE",
+        cache_dir / "composer-id-to-ws.json",
+    )
```
bradjin8, line 49: summary_cache_dir patches PROJECTS_CACHE_FILE and COMPOSER_MAP_CACHE_FILE but not tab-summary paths (_tab_summaries_path uses CACHE_DIR prefix only). Fine for current scope; add a comment that tab-summary benchmarks belong in issue #8's unified suite.

benchmarks/baselines.json
```diff
@@ -0,0 +1,15 @@
+{
+  "_note": "Gated means from ubuntu-latest CI benchmark-results.json (PR #120, run 28123677675). Refresh: pytest tests/benchmarks/ --benchmark-only --benchmark-json=benchmark-results.json -o addopts=",
```
bradjin8: _note refresh command is truncated (-o addopts= with no value). Complete the example: -o addopts= --benchmark-json=benchmark-results.json.

.github/workflows/tests.yml
```diff
@@ -213,3 +214,41 @@ jobs:
...
+          python -m pip install -r requirements-lock.txt
+          python -m pip install 'pytest>=8,<9' 'pytest-benchmark==4.0.0'
+
+      - name: Run summary-cache benchmarks
+        run: >
+          python -m pytest tests/benchmarks/
+          --benchmark-only
+          --benchmark-json=benchmark-results.json
+          --benchmark-columns=min,max,mean,stddev,rounds
+          -o addopts=
```
bradjin8, line 246: Consider pinning --benchmark-min-rounds (e.g. 5) to reduce CI timing variance near the 20% threshold, especially for sub-millisecond hit/miss benchmarks.

scripts/check_benchmark_regression.py
```diff
@@ -0,0 +1,163 @@
...
+            continue
+        if base == 0:
+            print(f"WARN: baseline for {name!r} is zero; skipping ratio check")
+            continue
+        ratio = cur / base
+        tag = "FAIL" if ratio > threshold else "ok"
+        print(f"[{tag}] {name}: {cur:.6f}s vs {base:.6f}s ({ratio:.2f}x)")
+        if ratio > threshold:
+            failures.append(name)
+
```
bradjin8, line 126: New benchmarks without baselines only WARN; gated missing benchmarks correctly FAIL (fixed in 83bb248). Consider also failing when all current results are faster than baseline by >50% (stale baseline drift), or document a baseline-refresh workflow in benchmarks/baselines.json.

bradjin8 (changes requested):

requirements-lock.txt
clean6378-max-it: Kept click==8.4.2 with an inline comment — lockfile CI failed on 8.4.1 vs pip-compile output; revert would re-break the lockfile job.

tests/benchmarks/test_summary_cache_bench.py
clean6378-max-it: Took the document path: module docstring clarifies miss = fingerprint mismatch on get_cached_projects, not rebuild. Merge deferred to #8 to avoid baseline name churn on this PR.

scripts/check_benchmark_regression.py
clean6378-max-it: Took the document path: _note in baselines.json now has the ubuntu refresh workflow. Skipped >50%-faster auto-fail — that would block intentional speedups until manual baseline refresh; can add in #8 if we want symmetric drift detection.

requirements-lock.txt
clean6378-max-it: the 8.4.1 → 8.4.2 bump stays (required by pip-compile on ubuntu); the extra comment can’t live in requirements-lock.txt because CI does a byte-for-byte line diff. Document the rationale in the PR reply instead.

bradjin8 (approved):

.github/workflows/tests.yml
```diff
@@ -213,3 +214,42 @@ jobs:
             --verbose \
             --redact \
             --exit-code 1
+
+  # ── Performance benchmarks: summary cache (issue #7) ───────────────────────
```
wpak-ai: issue #115 not #7

tests/benchmarks/conftest.py
```diff
@@ -0,0 +1,89 @@
...
+    return entries
+
+
+@pytest.fixture
+def summary_cache_dir(tmp_path: Path, monkeypatch: pytest.MonkeyPatch) -> Path:
+    """Redirect summary-cache files to an isolated temp directory.
+
+    Patches ``CACHE_DIR`` (also used by tab-summary paths via ``_tab_summaries_path``)
+    plus the projects/composer-map file constants used by current benchmarks.
+    Tab-summary cache benchmarks are deferred to issue #8 (unified benchmark suite).
```
wpak-ai: ditto

wpak-ai (approved):
Update the docstrings and you're ready to go.

PR #125, opened 2026-06-29
Cache invalid-workspace alias inference per storage fingerprint (#116)

clean6378-max-it:
## Summary
- Add mtime-keyed disk cache for `invalid_workspace_aliases` in `summary_cache.py`, using the same storage fingerprint invalidation flow as composer-map and tab-summary caches.
- Expose cached alias resolution on `WorkspaceContext` via `resolve_invalid_workspace_aliases_cached()` and `with_invalid_workspace_aliases()`.
- Update `assemble_single_tab`, tab summaries, and workspace listing to read precomputed aliases instead of re-scanning all `composerData:*` rows on every request.

## Problem
When `invalid_workspace_ids` is non-empty, hot paths such as `assemble_single_tab` ran a full global `composerData:*` scan on every deep-link request solely to build `invalid_workspace_aliases`. On large `state.vscdb` files this produced O(tabs × keys) I/O.

## Solution
Compute `invalid_workspace_aliases` once per storage fingerprint and serve subsequent requests from the disk cache until the fingerprint changes (DB replaced or storage mtimes change).

## Test plan
- [x] `pytest tests/test_workspace_tabs_summary.py tests/test_workspace_list_count_alignment.py -q`
- [x] Full `pytest -q` (532 passed, 21 skipped)
- [x] `mypy .`
- [x] Unit tests: cache hit, cache miss/invalidation, empty-alias case (all workspaces valid)
- [x] `assemble_single_tab` no longer runs global `composerData:%` scan on cache hit when invalid workspaces exist

## Out of scope
- Search index / live-scan paths (`search.py`) — deferred per operator
- Changing alias vote/tie-break semantics in `infer_invalid_workspace_aliases`

Closes #116

## Summary by CodeRabbit

* **New Features**
  * Added disk-backed caching for invalid workspace alias resolution to speed up workspace listings, tab summaries, and export flows.
  * Introduced a `nocache` option that bypasses the alias disk cache for workspace tab/listing and per-tab routes, and for export orchestration.

* **Bug Fixes**
  * Cached alias results are reused only when the workspace/storage fingerprint matches and are refreshed when global storage changes.
  * Malformed or corrupt cached alias data is safely ignored to prevent incorrect alias assignments.

### DIFF
--- a/services/summary_cache.py
+++ b/services/summary_cache.py
@@ -23,6 +23,7 @@ CACHE_VERSION = 1
 CACHE_DIR = Path.home() / ".cache" / "cursor-chat-browser"
 PROJECTS_CACHE_FILE = CACHE_DIR / "projects.json"
 COMPOSER_MAP_CACHE_FILE = CACHE_DIR / "composer-id-to-ws.json"
+INVALID_WORKSPACE_ALIASES_CACHE_FILE = CACHE_DIR / "invalid-workspace-aliases.json"
 TAB_SUMMARIES_PREFIX = "tab-summaries-"
 
 
@@ -234,6 +235,47 @@ def set_cached_composer_id_to_ws(
         {
             "fingerprint": fingerprint,
             "composer_id_to_ws": mapping,
+        },
+    )
+
+
+def get_cached_invalid_workspace_aliases(
+    fingerprint: dict[str, Any],
+) -> dict[str, str] | None:
+    """Load cached invalid-workspace alias map when the fingerprint matches.
+
+    Args:
+        fingerprint: Storage mtime/rules digest.
+
+    Returns:
+        ``{invalid_id: replacement_id}`` on hit, else ``None``.
+    """
+    data = _read_cache_file(INVALID_WORKSPACE_ALIASES_CACHE_FILE)
+    if not data:
+        return None
+    if not _fingerprint_equal(data.get("fingerprint"), fingerprint):
+        return None
+    aliases = data.get("invalid_workspace_aliases")
+    if not isinstance(aliases, dict):
+        return None
+    return {str(k): str(v) for k, v in aliases.items()}
+
+
+def set_cached_invalid_workspace_aliases(
+    fingerprint: dict[str, Any],
+    aliases: dict[str, str],
+) -> None:
+    """Persist invalid-workspace alias map under *fingerprint*.
+
+    Args:
+        fingerprint: Invalidation fingerprint paired with *aliases*.
+        aliases: ``{invalid_id: replacement_id}`` from alias inference.
+    """
+    _write_cache_file(
+        INVALID_WORKSPACE_ALIASES_CACHE_FILE,
+        {
+            "fingerprint": fingerprint,
+            "invalid_workspace_aliases": aliases,
         },
     )
 
--- a/services/summary_cache.py
+++ b/services/summary_cache.py
@@ -258,7 +258,12 @@ def get_cached_invalid_workspace_aliases(
     aliases = data.get("invalid_workspace_aliases")
     if not isinstance(aliases, dict):
         return None
-    return {str(k): str(v) for k, v in aliases.items()}
+    validated: dict[str, str] = {}
+    for key, value in aliases.items():
+        if not isinstance(key, str) or not isinstance(value, str):
+            return None
+        validated[key] = value
+    return validated
 
 
 def set_cached_invalid_workspace_aliases(
--- a/services/summary_cache.py
+++ b/services/summary_cache.py
@@ -257,10 +257,18 @@ def get_cached_invalid_workspace_aliases(
         return None
     aliases = data.get("invalid_workspace_aliases")
     if not isinstance(aliases, dict):
+        _logger.debug(
+            "Invalid workspace aliases cache rejected: invalid_workspace_aliases is not a dict",
+        )
         return None
     validated: dict[str, str] = {}
     for key, value in aliases.items():
         if not isinstance(key, str) or not isinstance(value, str):
+            _logger.debug(
+                "Invalid workspace aliases cache rejected: non-string entry (%r -> %r)",
+                key,
+                value,
+            )
             return None
         validated[key] = value
     return validated
--- a/services/workspace_context.py
+++ b/services/workspace_context.py
@@ -2,22 +2,27 @@
 
 from __future__ import annotations
 
+import os
 import sqlite3
-from dataclasses import dataclass, replace
+from dataclasses import dataclass, field, replace
 from typing import Any
 
 from models import Bubble
 from services.workspace_db import (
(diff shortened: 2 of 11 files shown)


services/workspace_tabs.py
```diff
@@ -582,22 +580,15 @@ def assemble_single_tab(
...
-                composer_rows=composer_rows_for_aliases,
-                project_layouts_map=project_layouts_map,
-                project_name_map=project_name_map,
-                workspace_path_map=workspace_path_map,
-                workspace_entries=workspace_entries,
-                bubble_map={},
-                composer_id_to_ws=composer_id_to_ws,
-                invalid_workspace_ids=invalid_workspace_ids,
-            )
+        invalid_workspace_aliases = resolve_invalid_workspace_aliases_cached(
```
bradjin8: assemble_single_tab does not pass nocache to resolve_invalid_workspace_aliases_cached (unlike listing/summary). Low impact today because GET .../tabs/<composer_id> has no ?nocache= param, but add the parameter now if you plan API parity with summary/list endpoints.

services/workspace_tabs.py
```diff
@@ -723,18 +714,13 @@ def assemble_workspace_tabs(
...
-                composer_rows=composer_rows,
-                project_layouts_map=project_layouts_map,
-                project_name_map=project_name_map,
-                workspace_path_map=workspace_path_map,
-                workspace_entries=workspace_entries,
-                bubble_map=bubble_map,
-                composer_id_to_ws=composer_id_to_ws,
-                invalid_workspace_ids=invalid_workspace_ids,
-            )
+        invalid_workspace_aliases = resolve_invalid_workspace_aliases_cached(
```
bradjin8: assemble_workspace_tabs also omits nocache on alias resolution. Full /tabs has no nocache flag today; fine unless you add one later.

services/workspace_context.py
```diff
@@ -135,3 +141,102 @@ def enrich_workspace_context_from_global_db(
...
+        bubble_map={},
+        composer_id_to_ws=ctx.composer_id_to_workspace_id,
+        invalid_workspace_ids=ctx.invalid_workspace_ids,
+    )
+    if not nocache_enabled(request_nocache=nocache):
+        set_cached_invalid_workspace_aliases(fingerprint, aliases)
+    return aliases
+
+
+def with_invalid_workspace_aliases(
```
bradjin8, line 222: with_invalid_workspace_aliases() is only used in tests. Call sites invoke resolve_invalid_workspace_aliases_cached directly and may re-read disk cache multiple times per request. Consider enriching ctx once per request in tab/list paths to avoid redundant cache file reads.

tests/test_workspace_listing_performance.py
```diff
@@ -126,6 +135,15 @@ def test_output_shape_preserved(self):
             self.assertIn("conversationCount", p)
             self.assertIn("lastModified", p)
 
+    def test_nocache_bypasses_alias_disk_cache(self):
+        ws_path = _make_fixture_with_invalid_workspace(self.tmp.name)
+        with patch(
+            "services.summary_cache.get_cached_invalid_workspace_aliases",
+        ) as mock_get:
+            mock_get.return_value = {"invalid-ws": "global"}
```
bradjin8, line 148: Nocache test patches only get_cached_invalid_workspace_aliases. Also assert set_cached_invalid_workspace_aliases is not called when nocache=True so write-bypass regressions are caught.

services/workspace_tabs.py
```diff
@@ -453,18 +455,14 @@ def _build_workspace_tab_summaries_uncached(
 
         composer_rows = safe_fetchall(global_db, COMPOSER_ROWS_WITH_HEADERS_SQL)
```
bradjin8, line 458: list_workspace_tab_summaries still fetches all composerData:* rows for the summary loop (required). Alias inference is cached across requests, but the first summary load per fingerprint still pays one roster scan — expected; worth a one-line comment so future readers don't try to "optimize" it away incorrectly.

services/summary_cache.py
```diff
@@ -238,6 +239,52 @@ def set_cached_composer_id_to_ws(
...
+    """
+    data = _read_cache_file(INVALID_WORKSPACE_ALIASES_CACHE_FILE)
+    if not data:
+        return None
+    if not _fingerprint_equal(data.get("fingerprint"), fingerprint):
+        return None
+    aliases = data.get("invalid_workspace_aliases")
+    if not isinstance(aliases, dict):
+        return None
+    validated: dict[str, str] = {}
```
bradjin8, line 264: Malformed cache entries reject the entire file (returns None). Correct and safe; consider logging at debug when validation fails so operators can distinguish corrupt cache from cold miss.

bradjin8 (changes requested):
## services/export_engine.py line 200:
Still calls infer_invalid_workspace_aliases inline on every export. Issue #116 scope lists export_engine as a call site; PR defers search but not export. Large exports with invalid workspaces will keep paying the full composer scan — consider resolve_invalid_workspace_aliases_cached here in a follow-up.

## services/search.py line 386:
Still uses inline infer_invalid_workspace_aliases in _load_search_workspace_assigner. PR marks search as out of scope; document in issue/PR that search deep-links may still cold-scan until a follow-up.

services/workspace_tabs.py
clean6378-max-it: assemble_single_tab(..., *, nocache: bool = False)
Passes nocache to resolve_workspace_context_cached and with_invalid_workspace_aliases
api/workspaces.py passes nocache=_request_nocache() on GET .../tabs/<composer_id>

services/workspace_tabs.py
clean6378-max-it: assemble_workspace_tabs(..., *, nocache: bool = False)
Passes nocache through context + alias resolution
api/workspaces.py passes nocache=_request_nocache() on full /tabs (not just ?summary=1)

services/workspace_context.py
clean6378-max-it: Tab paths now use with_invalid_workspace_aliases() instead of direct resolve_*

tests/test_workspace_listing_performance.py
clean6378-max-it: test_nocache_bypasses_alias_disk_cache patches both get_* and set_*; asserts neither is called.

services/workspace_tabs.py
clean6378-max-it: comment before composer_rows = safe_fetchall(...) in _build_workspace_tab_summaries_uncached

services/summary_cache.py
clean6378-max-it: Already fixed earlier (summary_cache.py 260-271 + test).

bradjin8 (approved):

wpak-ai (approved):

PR #133, opened 2026-07-15
Tighten Bubble boundary with typed accessors

clean6378-max-it:
Closes #128 

Public raw on Bubble, Composer, and WorkspaceLocalComposer is gone. 
Stored Cursor JSON now lives in _raw, so mypy --strict can enforce the domain boundary instead of letting callers reach straight into the dict.

Bubble accessors return TypedDict types from the new models/conversation_types.py (TokenCountDict, BubbleContextDict, and the rest) instead of dict[str, Any] and list[Any]. 
When the composers API needs to pass the full blob back to the client, it calls cursor_storage_payload() rather than copying .raw directly. raw_access.py bridge functions no longer take bare Any; 
they expect Bubble | dict or Composer | dict. Drift-warning behavior from PR #94 is unchanged.

Run mypy --strict . and pytest -q to verify. PR 2 stacks on this branch and routes the repetitive property bodies through _optional_dict_absent_ok / _optional_list_absent_ok.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consistency of composer/conversation payloads across per-workspace vs global lookups.
  * Preserved conversation data and related bubble content correctly when cursor payload formats vary.
  * Improved text extraction behavior for bubbles and tool-related content.

* **Reliability**
  * Strengthened validation via structured typing for conversation fields (metadata, context, model info, thinking, tool results).
  * Ensured cursor storage payloads are safely copied before use.
  * Improved tool parsing compatibility with read-only tool inputs.

* **Chores**
  * Updated dependency lock, benchmark baselines, and regression-gate filtering; refreshed related tests.

### DIFF
--- a/api/composers.py
+++ b/api/composers.py
@@ -201,7 +201,7 @@ def get_composer(composer_id: str) -> tuple[Response, int] | Response:
                             # whether it's absent or None, so the response shape
                             # is identical regardless of which branch resolved
                             # the composer (CodeRabbit on PR #30).
-                            payload = dict(local.raw)
+                            payload = local.cursor_storage_payload()
                             payload["conversation"] = payload.get("conversation") or []
                             return json_response(payload)
             except SchemaError as e:
@@ -240,7 +240,7 @@ def get_composer(composer_id: str) -> tuple[Response, int] | Response:
                             type(e).__name__,
                         )
                         return json_response({"error": "Composer schema drift"}, 404)
-                    payload = dict(composer.raw)
+                    payload = composer.cursor_storage_payload()
                     payload["conversation"] = payload.get("conversation") or []
                     return json_response(payload)
             except (OSError, sqlite3.Error, json.JSONDecodeError, ValueError):
--- a/api/composers.py
+++ b/api/composers.py
@@ -98,12 +98,13 @@ def list_composers() -> tuple[Response, int] | Response:
                         # load-bearing, not just a filter (Brad's review): the
                         # sort key and the JSON's composerId both read off the
                         # validated values, not the raw dict.
-                        c["composerId"] = local.composer_id
-                        c["lastUpdatedAt"] = local.last_updated_at
-                        c["conversation"] = c.get("conversation") or []
-                        c["workspaceId"] = name
-                        c["workspaceFolder"] = workspace_folder
-                        composers.append((local, c))
+                        payload = local.cursor_storage_payload()
+                        payload["composerId"] = local.composer_id
+                        payload["lastUpdatedAt"] = local.last_updated_at
+                        payload["conversation"] = payload.get("conversation") or []
+                        payload["workspaceId"] = name
+                        payload["workspaceFolder"] = workspace_folder
+                        composers.append((local, payload))
             except SchemaError as e:
                 _logger.warning(
                     "Schema drift in %s: %s (%s)",
--- a/models/conversation.py
+++ b/models/conversation.py
@@ -27,6 +27,50 @@ from models.from_dict_validation import (
 _logger = logging.getLogger(__name__)
 
 
+def _filter_str_list_elements(
+    value: list[Any],
+    *,
+    model: str,
+    record_id: str,
+    field: str,
+) -> list[str]:
+    filtered: list[str] = []
+    for item in value:
+        if isinstance(item, str):
+            filtered.append(item)
+        else:
+            _logger.warning(
+                "Schema drift in %s %s: invalid %s element (expected str, got %s)",
+                model,
+                record_id,
+                field,
+                type(item).__name__,
+            )
+    return filtered
+
+
+def _filter_dict_list_elements(
+    value: list[Any],
+    *,
+    model: str,
+    record_id: str,
+    field: str,
+) -> list[FileUriDict]:
+    filtered: list[FileUriDict] = []
+    for item in value:
+        if isinstance(item, dict):
+            filtered.append(cast(FileUriDict, item))
+        else:
+            _logger.warning(
+                "Schema drift in %s %s: invalid %s element (expected dict, got %s)",
+                model,
+                record_id,
+                field,
+                type(item).__name__,
+            )
+    return filtered
+
+
 @dataclass(frozen=True)
 class Composer:
     """Cursor conversation row from globalStorage cursorDiskKV; requires fullConversationHeadersOnly + createdAt."""
@@ -287,7 +331,12 @@ class Bubble:
                 type(value).__name__,
             )
             return []
-        return cast(list[str], value)
+        return _filter_str_list_elements(
+            value,
+            model="Bubble",
+            record_id=self.bubble_id,
+            field="relevantFiles",
+        )
 
     @property
     def attached_file_code_chunks_uris(self) -> list[FileUriDict]:
@@ -301,7 +350,12 @@ class Bubble:
                 type(value).__name__,
             )
             return []
-        return cast(list[FileUriDict], value)
+        return _filter_dict_list_elements(
+            value,
+            model="Bubble",
+            record_id=self.bubble_id,
+            field="attachedFileCodeChunksUris",
+        )
 
     @property
     def context(self) -> BubbleContextDict:
--- a/models/conversation.py
+++ b/models/conversation.py
@@ -4,6 +4,17 @@ import logging
 from dataclasses import dataclass, field
 from typing import Any, cast
 
+from models.conversation_types import (
+    BubbleContextDict,
+    BubbleMetadataDict,
+    ContextWindowStatusDict,
+    FileUriDict,
+    ModelInfoDict,
+    ThinkingDict,
+    TokenCountDict,
+    ToolFormerDataDict,
+    ToolResultEntry,
+)
 from models.errors import SchemaError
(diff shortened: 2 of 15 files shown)


api/composers.py
bradjin8, line 1: Line 101

`get_composer` now builds the response from `local.cursor_storage_payload()` (lines 204–205), but `list_composers` still mutates `c` in place. Because `WorkspaceLocalComposer.from_dict(c)` stores `_raw=c` by reference, these writes also mutate the model’s internal storage and inject response-only keys (`workspaceId`, `workspaceFolder`) into `_raw`.

Issue #128 AC: “Direct `.raw[...]` reads outside model removed or funneled.” Please mirror the single-composer path:

```python
payload = local.cursor_storage_payload()
payload["composerId"] = local.composer_id
payload["lastUpdatedAt"] = local.last_updated_at
payload["conversation"] = payload.get("conversation") or []
payload["workspaceId"] = name
payload["workspaceFolder"] = workspace_folder
composers.append((local, payload))

File: models/conversation.py [hunk 1]
```diff
@@ -258,11 +287,11 @@ def relevant_files(self) -> list[Any]:
                 type(value).__name__,
             )
             return []
-        return value
+        return cast(list[str], value)
```
bradjin8: `relevant_files` now advertises `list[str]`, but after the list-type check the code does `return cast(list[str], value)` with no per-element validation. If Cursor ever drifts to URI objects (like `newlyCreatedFiles` / `attachedFileCodeChunksUris`), callers such as `workspace_resolver.get_project_from_file_path(fp, …)` will receive non-strings at runtime while mypy thinks they are safe.

Consider filtering or warn-don’t-coerce on non-`str` elements (same philosophy as PR #94), e.g. skip bad entries with a schema-drift warning rather than a blind cast.

tests/test_blob_parsing_fuzz.py
```diff
@@ -184,7 +184,7 @@ def _assemble_workspace_bubble(bubble_id: object, value: object) -> dict | None:
     except (json.JSONDecodeError, TypeError, ValueError):
         return None
     try:
-        return Bubble.from_dict(parsed, bubble_id=bubble_id).raw  # type: ignore[arg-type]
+        return Bubble.from_dict(parsed, bubble_id=bubble_id)._raw  # type: ignore[arg-type]
```
bradjin8: This helper reaches through the new private boundary via `Bubble.from_dict(...)._raw`. For fuzz/prod parity with the API boundary introduced in this PR, prefer `cursor_storage_payload()` unless the test explicitly needs reference identity to the parsed dict (in which case a one-line comment explaining why `_raw` is required would help future readers).

models/conversation.py
```diff
@@ -272,11 +301,11 @@ def attached_file_code_chunks_uris(self) -> list[Any]:
                 type(value).__name__,
             )
             return []
-        return value
+        return cast(list[FileUriDict], value)
```
bradjin8: Same pattern as `relevant_files`: `attached_file_code_chunks_uris` casts the whole list to `list[FileUriDict]` without checking elements are dicts. A wrong element type would slip through silently. Low urgency if live payloads are stable, but element-level drift warnings would match the rest of the model layer.

models/conversation_types.py
```diff
@@ -0,0 +1,61 @@
...
+from typing import Any, TypedDict
+
+
+class FileUriDict(TypedDict, total=False):
+    """URI object on attached-file or context entries."""
+
+    path: str
+
+
+class BubbleMetadataDict(TypedDict, total=False):
```
bradjin8, line 14: `BubbleMetadataDict` currently documents only `modelName`. Cursor’s storage `metadata` blob also carries token counts, tool calls, etc. (see `build_storage_bubble_metadata` / tab aggregation). Either extend the TypedDict with the fields we actually read, or add a module comment that this is intentionally minimal and display metadata is assembled elsewhere — avoids misleading future callers into thinking `bubble.metadata` is a complete shape.

bradjin8 (changes requested):

api/composers.py
clean6378-max-it: good catch. list_composers now uses cursor_storage_payload() like get_composer, so we enrich a copy instead of mutating _raw.

models/conversation.py
clean6378-max-it: yeaah , dropped the  blind cast . no-str-entires get a drift warning and are skipped.
teste added in test_raw_accessors.py

tests/test_blob_parsing_fuzz.py
clean6378-max-it: helper now returns cursor_storage_payload().
the assesrtIs(bubble._raw, raw) test stays so it check that from dict keeps the parsed dict by reference

models/conversation.py
clean6378-max-it: bad elements skipped with a warning. covered in test_raw_accessors.py.

models/conversation_types.py
clean6378-max-it: added a doc note

bradjin8 (approved):

models/raw_access.py
bradjin8, line 1: ## Optional, nice to have
`Bubble.relevant_files` now filters non-`str` elements with drift warnings, but the dict fallback branch here still returns the raw list via `_optional_list_absent_ok` + cast. Production currently always passes `Bubble`, so this isn't hit today — but the bridge contract is inconsistent. Consider reusing `_filter_str_list_elements` (or routing dict inputs through the same filter) so both paths behave identically.

models/conversation.py
```diff
@@ -384,12 +467,12 @@ def tool_results(self) -> list[Any] | None:
                 type(value).__name__,
             )
             return None
-        return value
+        return cast(list[ToolResultEntry], value)
```
bradjin8, line 470: `tool_results` still does a blind `cast(list[ToolResultEntry], value)` with no per-element validation. Same pattern you just fixed for `relevantFiles` and `attachedFileCodeChunksUris`. Low urgency if live payloads are stable; could follow up in PR 2.

api/composers.py
```diff
@@ -98,12 +98,13 @@ def list_composers() -> tuple[Response, int] | Response:
                         # load-bearing, not just a filter (Brad's review): the
                         # sort key and the JSON's composerId both read off the
                         # validated values, not the raw dict.
-                        c["composerId"] = local.composer_id
-                        c["lastUpdatedAt"] = local.last_updated_at
-                        c["conversation"] = c.get("conversation") or []
-                        c["workspaceId"] = name
-                        c["workspaceFolder"] = workspace_folder
-                        composers.append((local, c))
+                        payload = local.cursor_storage_payload()
```
bradjin8, line 101: Fix looks correct
response fields are written to the shallow copy from `cursor_storage_payload()`, not the SQLite-parsed dict aliased as `_raw`. Consider a small wired test (mutate list response, assert `workspaceId` absent from `local._raw`) to pin this boundary against regressions.

bradjin8 (approved):

models/raw_access.py
clean6378-max-it: dict fallback now goes through the same filter helpers as the bubble property. 
test added in test_raw_accessors.py.

models/conversation.py
clean6378-max-it: agreed, same pattern,  will handle in next pr with the raw-access dedup work

api/composers.py
clean6378-max-it: alright, added test_list_composers_response_fields_do_not_mutate_storage_raw to pin that boundary

tests/test_raw_accessors.py
wpak-ai, line 1: add a bubble_context({...}, "id") test to cover the dict-bridge path. The sibling bridges (bubble_relevant_files, bubble_attached_file_uris) are already tested this way.

models/raw_access.py
```diff
@@ -266,47 +268,55 @@ def composer_code_block_data(data: Any, composer_id: str) -> dict[str, Any] | No
...
-        )
-        return ctx if ctx is not None else {}
-    return {}
+    ctx = _optional_dict_absent_ok(
+        bubble,
+        "context",
+        model="Bubble",
+        entity_id=bubble_id,
+    )
+    return cast(BubbleContextDict, ctx if ctx is not None else {})
```
wpak-ai, line 324: cast()'s first argument is evaluated at runtime, so the dict-bridge branch raises NameError. mypy --strict misses it (it treats TYPE_CHECKING as true) and no test exercises the branch, so CI is green.
Fix — give it a runtime import (safe; conversation_types.py only imports typing, so no circular import):
```
from models.conversation_types import BubbleContextDict  # module level, not under TYPE_CHECKING
```

wpak-ai (changes requested):

tests/test_raw_accessors.py
clean6378-max-it: fixed, BubbleContextDict is a runtime import now. reproduced the NameError on the dict path before the fix

models/raw_access.py
clean6378-max-it: added dict-bridge tests for bubble_context. same pattern as the other bridges

wpak-ai (approved):

PR #134, opened 2026-07-15
types: thread RuleTokens through exclusion-rules consumers

clean6378-max-it:
Closes #130 

utils/exclusion_rules.py already defines RuleTokens and uses it on load_rules and is_excluded_by_rules. the callers did not. search, workspace tabs, export, summary cache, and the other services that pass rules around still typed them as list[Any]. flask_config.exclusion_rules() returned list[list[Any]]. with mypy strict, that gap sits right where sensitive sessions get filtered.

this pr is annotations only. each consumer imports RuleTokens from utils.exclusion_rules and takes list[RuleTokens] instead of list[Any]. flask_config.exclusion_rules() now returns list[RuleTokens]. no runtime logic changes.

to verify, run mypy --strict services/ api/flask_config.py utils/exclusion_rules.py, then pytest tests/ -q -k exclu and the full test suite.

also bumps pillow in requirements-lock.txt (12.2.0 → 12.3.0) so lock file freshness passes; unrelated to RuleTokens, stale pin only.

## Summary by CodeRabbit

- **Refactor**
  - Tightened exclusion-rule typing across search, workspace tabs/listing, exports, caching, indexing, and workspace composition/scanning to use a shared, token-based rule shape.
  - Updated the search API route and app initialization to source and reuse exclusion rules consistently during request handling and background index startup.

- **Chores**
  - Updated the pinned `pillow` version in the lockfile (12.2.0 → 12.3.0).

### DIFF
--- a/api/flask_config.py
+++ b/api/flask_config.py
@@ -6,8 +6,10 @@ from typing import Any, overload
 
 from flask import Response, current_app, jsonify
 
+from utils.exclusion_rules import RuleTokens
 
-def exclusion_rules() -> list[list[Any]]:
+
+def exclusion_rules() -> list[RuleTokens]:
     """Return loaded exclusion rules from app config (empty list when unset)."""
     return current_app.config.get("EXCLUSION_RULES") or []
 
--- a/api/flask_config.py
+++ b/api/flask_config.py
@@ -2,7 +2,7 @@
 
 from __future__ import annotations
 
-from typing import Any, overload
+from typing import Any, cast, overload
 
 from flask import Response, current_app, jsonify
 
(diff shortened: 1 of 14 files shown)


bradjin8 (changes requested):
## api/search.py:141
`workspaces.py` and `export_api.py` both obtain rules via `exclusion_rules()` from `api/flask_config`, but search still reads `current_app.config.get("EXCLUSION_RULES")` directly. That bypasses the typed producer this PR introduces and leaves the last API-layer gap in the #130 “thread RuleTokens end-to-end” goal.

```python
from api.flask_config import exclusion_rules, json_response
# ...
rules = exclusion_rules()
```

## app.py:78
Startup passes `app.config["EXCLUSION_RULES"]` (untyped `Any` from Flask config) into `start_search_index_background`, which now expects `list[RuleTokens]`. Mypy passes via `Any`, but for consistency with `flask_config.exclusion_rules()` you could assign once at startup:
```python
loaded_rules = load_rules(resolved)
app.config["EXCLUSION_RULES"] = loaded_rules
# ...
start_search_index_background(resolve_workspace_path(), loaded_rules)

bradjin8 (approved):

wpak-ai (approved):

PR #137, opened 2026-07-16
dedupe bubble/composer reads through raw_access helpers

clean6378-max-it:
CLoses #129 

follow-up to #133. bubble and composer properties were all doing the same optional-field dance (get key, warn on drift, return a default). this routes them through the shared helpers in raw_access.py instead.

behavior should be unchanged: same return values, same warning text. tests in test_raw_accessors.py cover wrong types on the model path and the dict-bridge path.

sanity check: pytest -k "bubble or composer or raw_access" and mypy ..

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved robustness when loading saved conversations and messages with missing or unexpectedly formatted fields.
  * More consistent safe fallbacks for metadata, attached files, context, token counts, tool results, usage data, and related fields.
  * Enhanced diagnostics for malformed saved data so problematic fields are easier to identify during troubleshooting.

### DIFF
--- a/models/conversation.py
+++ b/models/conversation.py
@@ -376,6 +376,7 @@ class Bubble:
 
     @property
     def thinking(self) -> str | ThinkingDict | None:
+        # Inline: accepts str | dict; no raw_access helper for that union.
         value = self._raw.get("thinking")
         if value is None:
             return None
@@ -390,6 +391,7 @@ class Bubble:
 
     @property
     def thinking_duration_ms(self) -> int | float | None:
+        # Inline: absent -> None (not 0); bool guard before numeric check.
         value = self._raw.get("thinkingDurationMs")
         if value is None:
             return None
--- a/models/conversation.py
+++ b/models/conversation.py
@@ -23,6 +23,13 @@ from models.from_dict_validation import (
     require_non_empty_str_field,
     require_type,
 )
+from models.raw_access import (
+    _optional_dict_absent_ok,
+    _optional_dict_default_empty,
+    _optional_list_absent_none,
+    _optional_list_absent_ok,
+    _optional_number_absent_ok,
+)
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


models/conversation.py
```diff
@@ -307,111 +292,87 @@ def text(self) -> str | None:
...
+            _optional_dict_default_empty(
+                self._raw,
+                "modelInfo",
+                model="Bubble",
+                entity_id=self.bubble_id,
+            ),
+        )
 
     @property
     def thinking(self) -> str | ThinkingDict | None:
```
bradjin8, line 378: `thinking` and `thinking_duration_ms` still use inline drift handling while every other optional Bubble field routes through `raw_access`. That's reasonable given the `str | dict` union, but a one-line comment on each property explaining why they stay inline would help future readers avoid "finish the migration" churn.

models/raw_access.py
```diff
@@ -154,6 +154,68 @@ def _optional_dict_absent_ok(
     return value
 
 
+def _optional_dict_default_empty(
+    raw: dict[str, Any],
```
bradjin8, line 164: `_optional_dict_default_empty`, `_optional_list_absent_none`, and `_optional_number_absent_ok` are imported across the package boundary from `conversation.py` while still underscore-prefixed. Consider either (a) dropping the leading underscore now that they're shared implementation, or (b) a brief module doc note that these are package-internal shared helpers - makes the intended API surface explicit.

models/conversation.py
bradjin8, line 1: Line 37: 
`_filter_str_list_elements` / `_filter_dict_list_elements` live here but are lazy-imported by `raw_access.bubble_*` bridge functions, creating a conversation ↔ raw_access cycle. Works today; if this grows, moving the filters into `raw_access` (or a small `models/raw_list.py`) would break the cycle without changing behavior.

bradjin8 (changes requested):

models/conversation.py
clean6378-max-it: added a one-line note on each: str|dict union on thinking, absent→none + bool guard on duration. no helper for either yet.

models/raw_access.py
clean6378-max-it: agreed, extended the raw_access module doc
those helpers are package-internal shared code , not single-module private, remaining can be a follow-up

models/conversation.py
clean6378-max-it: yeah lazy imports keep it safe for now

bradjin8 (approved):

wpak-ai (approved):

PR #138, opened 2026-07-16
cleanup workspace_tabs and cursor_md_exporter hygiene debt

clean6378-max-it:
Closes #136 

follow-up from the trim-model-exports review. broke up the big tab assembly and markdown export paths into smaller helpers, pulled the duplicate composer loop and tool-call rendering into shared functions, and replaced a few broad except Exception blocks with narrower ones. display_bubble.py now returns BubbleMetadata | None from build_storage_bubble_metadata so the casts can go.

no intentional user-visible behavior change. mypy --strict on the three files and full pytest -q (569 passed).

## Summary by CodeRabbit

- **New Features**
  - Workspace tabs now more reliably assemble per-bubble context and can include context-token usage/limits metadata when available.
  - Cursor Markdown exports provide clearer session summaries, improved activity/tool-call formatting, and optional collapsible “Thinking” details.
- **Bug Fixes**
  - More consistent messaging when global storage or conversations are missing.
  - Prevents malformed composer/token-count data from breaking tab listing and hardens CLI metadata parsing.
- **Tests**
  - Added regression tests for tab metadata aggregation and malformed workspace composer handling.
  - Added markdown exporter tests for non-dict metadata and web-only tool stats summaries.

### DIFF
--- a/services/workspace_tabs.py
+++ b/services/workspace_tabs.py
@@ -5,9 +5,9 @@ import json
 import logging
 import os
 import sqlite3
-from collections.abc import Mapping
+from collections.abc import Iterator, Mapping
 from datetime import datetime
-from typing import Any, cast
+from typing import Any
 
 _logger = logging.getLogger(__name__)
 
@@ -19,6 +19,7 @@ from utils.path_helpers import (
 )
 from utils.exclusion_rules import RuleTokens, build_searchable_text, is_excluded_by_rules
 from utils.display_bubble import (
+    annotate_response_times,
     bubble_display_timestamp_ms,
     build_storage_bubble_metadata,
 )
@@ -29,9 +30,8 @@ from models import (
     Composer,
     DisplayBubble,
     ParseWarningCollector,
-    SchemaError,
 )
-from models.bubble_display import BubbleMetadata, BubbleRole
+from models.bubble_display import BubbleRole
 from models.raw_access import (
     conversation_header_bubble_id,
     message_request_context_project_layouts,
@@ -73,6 +73,10 @@ from services.workspace_resolver import (
     matching_workspace_ids_for_folder,
 )
 
+
+ERR_CONVERSATION_NOT_FOUND = "Conversation not found"
+ERR_GLOBAL_STORAGE_NOT_FOUND = "Global storage not found"
+_COMPOSER_ROW_ERRORS = (KeyError, TypeError, ValueError, json.JSONDecodeError)
 
 
 def _loads_kv_value_logged(key: str, raw: object | None) -> Any | None:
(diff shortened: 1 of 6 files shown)


services/workspace_tabs.py
bradjin8, line 1: ## _aggregate_tab_metadata still annotated dict[str, Any] | None while building typed fields
could return BubbleMetadata-shaped dict for consistency with display_bubble.py.

bradjin8 (approved):
No dedicated regression tests for the three 5197228 edge cases (contextTokenLimit-only metadata, non-dict CLI meta JSON, web-only tool_result_stats summary). Existing test_workspace_tabs_* / test_cursor_md_exporter.py pass but don’t pin these paths.

services/workspace_tabs.py
clean6378-max-it: i didn't follow you,  didn't change the return type. still dict[str, Any] | None.

BubbleMetadata is per-bubble (modelName, inputTokens, etc.). _aggregate_tab_metadata builds tab totals (totalInputTokens, modelsUsed, ...). wrong shape to reuse it.

did fix a related bug in that function in commit 5197228: has_any now counts ctx_token_limit, so limit-only metadata isn't dropped.

TabMetadata typedict would be a follow-up if we want stricter typing here. out of scope for this pr unless you want it in.

services/workspace_tabs.py
```diff
@@ -734,52 +799,32 @@ def assemble_workspace_tabs(
...
+            project_name_map=project_name_map,
+            workspace_path_map=workspace_path_map,
+            workspace_entries=workspace_entries,
+            composer_id_to_ws=composer_id_to_ws,
+            invalid_workspace_ids=invalid_workspace_ids,
+            invalid_workspace_aliases=invalid_workspace_aliases,
+            matching_ws_ids=matching_ws_ids,
+            bubble_map={},
+        ):
+            tab = _assemble_tab_from_composer_data(
```
wpak-ai: assemble_workspace_tabs now calls _assemble_tab_from_composer_data(...) bare in the consumer loop. The base code wrapped each per-composer body in except Exception (log + record + skip); the new generator's except _COMPOSER_ROW_ERRORS only guards assign_composer_workspace, and a for-body exception is never routed back into a generator. One bubble with nested-drifted metadata (e.g. tokenCount={"inputTokens": "1500"}) raises TypeError at an unguarded if in_tok > 0 in the shared build_storage_bubble_metadata and 500s the whole listing.

wpak-ai (changes requested):

services/workspace_tabs.py
clean6378-max-it: added _safe_assemble_tab_from_composer_data so drift in one composer logs, records the failure, and skips instead of 500ing the listing. regression test covers string inputTokens.

wpak-ai (approved):

PR #147, opened 2026-07-21
feat: add Content-Security-Policy header with nonce for inline bootstrap

clean6378-max-it:
Closes #141 

adds a content-security-policy header on html pages only. each response gets a random nonce; inline scripts in the page templates use it, and the head theme bootstrap moved to static/js/theme-bootstrap.js so script-src stays tight. json api routes don't get the header.

pulled all onclick/onchange handlers out of the templates and wired them with addeventlistener instead, including sidebar click delegation in workspace.html. policy allows self plus cdnjs for scripts and styles; style-src keeps unsafe-inline because highlight.js themes need it.

tests/test_csp_header.py checks the header on page routes, nonce matches the inline script, and /api/workspaces has no csp. pytest tests/test_csp_header.py passes.

## Summary by CodeRabbit

- **Security**
  - Enabled per-response Content Security Policy (CSP) with nonce protection for all HTML pages, and added tests to verify nonce matching, correct directive behavior, and omission of CSP on JSON endpoints.
- **User Experience**
  - Improved theme initialization to apply the stored theme earlier.
  - Removed inline event handlers across the UI (theme toggle, search filters/actions, config save, exports, workspace actions) and replaced them with script-driven wiring.
- **Bug Fixes**
  - Improved “Copy all” button detection for more reliable clipboard behavior.

### DIFF
--- a/app.py
+++ b/app.py
@@ -6,12 +6,13 @@ from the Cursor editor's AI chat feature.
 
 import logging
 import os
+import secrets
 import sys
 from datetime import datetime
 from pathlib import Path
 from typing import cast
 
-from flask import Flask, Response, render_template, send_from_directory
+from flask import Flask, Response, g, render_template, send_from_directory
 
 from utils.debug_flag import resolve_debug_flag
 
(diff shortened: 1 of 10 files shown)


bradjin8 (approved):
**Nice to have: add `/config` to CSP header tests**

`test_html_page_has_content_security_policy_header` covers `/` and workspace covers `/workspace/global`, but `/config` is the only remaining page route without an explicit CSP assertion. Quick addition for full page-route coverage.

wpak-ai (approved):

PR #150, opened 2026-07-21
fix: return 409 from set-workspace when multiple WSGI workers are in use

clean6378-max-it:
Closes #144 

With gunicorn running more than one worker, POST set-workspace only changed the in-memory override on the worker that handled the request. 
The response still looked like success, so other workers kept the old path and operators had no signal that the change did not stick fleet-wide.

This branch blocks that case. 
After path validation, the handler checks whether the deployment looks multi-worker using environment signals such as WEB_CONCURRENCY, GUNICORN_WORKERS, worker flags in GUNICORN_CMD_ARGS, or an explicit CURSOR_BROWSER_MULTI_WORKER override. 
When that check is true, the API returns HTTP 409 with error code set_workspace_multi_worker_unsupported and tells the operator to set WORKSPACE_PATH or pass base-dir at startup instead of using the Configuration page. Single-process and single-worker behavior is unchanged when those signals are absent.

DEPLOYMENT.md now describes the 409 response and the env vars involved. tests/test_set_workspace_multiworker.py covers the 409 path, the success path when detection is off, and the env parsing helpers. 
Sanity check by running that test module and the existing set-workspace API tests in test_workspace_path_validation.py.

## Summary by CodeRabbit

* **Bug Fixes**
  * Workspace updates now return a clear HTTP **409** when multiple worker processes are detected, and the workspace change is blocked.
  * Configuration page now shows an actionable error message (no redirect) when the workspace update fails, including the **409** case.
  * Single-process deployments still allow workspace changes.

* **Documentation**
  * Updated production and multi-process deployment guidance, including worker settings and **409** behavior for `POST /api/set-workspace`.

* **Tests**
  * Added/expanded tests covering multi-worker detection and the `/api/set-workspace` **409** vs **200** outcomes.

### DIFF
--- a/DEPLOYMENT.md
+++ b/DEPLOYMENT.md
@@ -66,10 +66,10 @@ gunicorn and waitress are **not** runtime dependencies; install them only when d
 
 When gunicorn runs with `--workers 2` (or more), each worker is an independent Python process:
 
-- **`POST /api/set-workspace`** only updates the worker that handled that request. Other workers keep their previous override (usually `None`). Prefer setting `WORKSPACE_PATH` or passing `--base-dir` at process start so every worker sees the same path.
+- **`POST /api/set-workspace`** cannot update every worker from a single request (each process has its own override). The API returns **HTTP 409** with code `set_workspace_multi_worker_unsupported` instead of success when multiple workers are detected (`WEB_CONCURRENCY`, `GUNICORN_WORKERS`, gunicorn `--workers` in `GUNICORN_CMD_ARGS`, or `CURSOR_BROWSER_MULTI_WORKER=1`). Set `WORKSPACE_PATH` or pass `--base-dir` at process start so every worker sees the same path.
 - **Exclusion rules** (`EXCLUSION_RULES` in app config) are loaded once at worker startup from `--exclude-rules` or the default file. Changing the rules file requires restarting workers.
 
-For a single-user localhost deployment with multiple workers, set the workspace path via environment variable rather than the Configuration page.
+For a single-user localhost deployment with multiple workers, set the workspace path via environment variable or `--base-dir` at startup; do not rely on the Configuration page `set-workspace` call.
 
 ## Path configuration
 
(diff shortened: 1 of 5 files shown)


DEPLOYMENT.md
bradjin8, line 1: **Should fix: align DEPLOYMENT.md gunicorn example with multi-worker detection**

The example at DEPLOYMENT.md:15 (`gunicorn ... --workers 2 ...`) doesn't set any env var that `is_multi_worker_process_deployment()` reads. Bare gunicorn workers usually won't have `WEB_CONCURRENCY` / `GUNICORN_CMD_ARGS` set, so POST /api/set-workspace may still return 200.

Please either:
- Prefix the example with `WEB_CONCURRENCY=2` (or `CURSOR_BROWSER_MULTI_WORKER=1`), or
- Add a note that multi-worker gunicorn deployments must export one of the documented detection env vars.

This closes the gap between the documented deployment command and issue #144's manual verification step.

tests/test_set_workspace_multiworker.py
```diff
@@ -0,0 +1,88 @@
...
+            return_value=True,
+        ):
+            resp = self.client.post(
+                "/api/set-workspace",
+                json={"path": self.storage},
+            )
+        self.assertEqual(resp.status_code, 409)
+        body = resp.get_json()
+        self.assertEqual(body["code"], "set_workspace_multi_worker_unsupported")
+        self.assertIn("WORKSPACE_PATH", body["error"])
```
bradjin8, line 43: test_multi_worker_returns_409_with_stable_code could assert get_workspace_path_override() is still None after 409, proving we don't partially apply the change.

bradjin8 (changes requested):
**Nice to have: handle 409 in config page**

`config.html` calls `/api/set-workspace` without checking the response status. On 409, the UI may still show success. Consider surfacing `body.error` / `body.code` when `res.status === 409`.

DEPLOYMENT.md
clean6378-max-it: Updated the multi-worker quick start to WEB_CONCURRENCY=2 gunicorn ... --workers 2 with a short note on why, so the example matches what detection reads.

tests/test_set_workspace_multiworker.py
clean6378-max-it: Done. The test now asserts get_workspace_path_override() is still None after the 409.

bradjin8 (approved):

wpak-ai (approved):

PR #151, opened 2026-07-21
playwright xss tests for renderMarkdownSafe

clean6378-max-it:
Closes #142 

The grep tests in test_xss_sanitization.py only prove the source still mentions DOMPurify. they do not run the frontend. this adds a headless chromium test that loads a real page (marked + dompurify + app.js from base.html), calls renderMarkdownSafe, and drops the result into innerHTML the same way workspace does.

Four payloads from the issue: img onerror, script tag, javascript: link markdown, svg onload. we set window.__xssProbe and check the dom sink for leftover onerror, script nodes, javascript: hrefs, and svg onload. production pages ship with csp now, so inline handlers often will not fire even if markup slips through. the assertions target stripped markup, not alert dialogs. there is a negative control that runs bare marked.parse without sanitize and expects dangerous attributes to still show up in the tree.

playwright 1.50.0 is in dev deps. ci gets a ubuntu-only browser-xss job: install chromium with deps, pytest tests/test_xss_browser.py, then the existing unittest grep backstop. pyinstaller steps stay on the unittest matrix job.

locally: pip install -e ".[dev]", playwright install chromium, pytest tests/test_xss_browser.py. comment out DOMPurify.sanitize in renderMarkdownSafe and the safe-path tests should fail.

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened confidence in Markdown XSS protection via end-to-end browser checks that confirm unsafe markup and links are blocked, with a negative-control scenario to detect regressions.
* **Tests**
  * Added Playwright-based headless browser regression tests for the production Markdown rendering flow.
  * Updated static sanitization verification guidance to align with the new browser coverage.
* **Chores**
  * Updated CI to run the new browser XSS checks on Ubuntu using a pinned Playwright version.

### DIFF
--- a/tests/test_xss_browser.py
+++ b/tests/test_xss_browser.py
@@ -1,0 +1,163 @@
+"""
+Headless-browser XSS regression tests (issue #11 / sprint item #3).
+
+Exercises the production render path: Marked.js + DOMPurify via
+renderMarkdownSafe() in static/js/app.js, then DOM insertion via innerHTML
+(the same pattern as templates/workspace.html).
+
+Pages are served with the same CSP as production; inline event handlers are
+blocked by the header, so these tests assert dangerous markup is stripped from
+the sink (and use a probe where execution is still observable).
+
+Static source greps live in tests/test_xss_sanitization.py.
+
+Run:
+    playwright install chromium
+    pytest -q tests/test_xss_browser.py
+"""
+
+from __future__ import annotations
+
+import threading
+from typing import Any, Generator
+
+import pytest
+from playwright.sync_api import Page, sync_playwright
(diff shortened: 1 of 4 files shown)


tests/test_xss_browser.py
bradjin8, line 1: Optional data: URI vector (app.js mentions it)

bradjin8 (approved):

tests/test_xss_browser.py
clean6378-max-it: Added a data_uri param case ([x](data:text/html,<script>...)) and assert no href^="data:" in the sink after renderMarkdownSafe.

wpak-ai (approved):

PR #158, opened 2026-07-29
fix: serialise summary-cache fingerprint read-compare-write under a module lock

clean6378-max-it:
Closes #154 
Under threaded WSGI, two requests could both miss the summary cache, build in parallel, and the slower one would overwrite the faster one's write. That's the bug.

This PR adds a module lock in summary_cache.py and get_or_build_cached_* helpers for all four cache types. Fingerprint and cache lookup happen under the lock; the build runs outside it; then we re-fingerprint and recheck before writing. If another thread filled the cache while we were building, we return that result and skip the write.

The four call sites (workspace_listing, workspace_tabs, workspace_db, workspace_context) now go through those helpers. tests/test_summary_cache_concurrency.py has a deterministic lost-update test plus a warm-cache stress test. Full suite: 614 passed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of workspace summary caching under concurrent access by serializing cache fingerprinting and cache read/update operations.
  * Prevented lost-update races with double-checked “recheck before write” cache rebuild flow.
  * Ensured workspace tab summaries are cached only when the underlying request succeeds (HTTP 200).
  * Extended consistent caching behavior across workspace project listings, composer/workspace mappings, and invalid workspace alias resolution.
* **Tests**
  * Updated concurrency regression coverage to use the new public cache fingerprint helper and validate consistent results during blocked rebuilds.

### DIFF
--- a/services/summary_cache.py
+++ b/services/summary_cache.py
@@ -17,7 +17,7 @@ import threading
 from collections.abc import Callable
 from concurrent.futures import ThreadPoolExecutor, as_completed
 from pathlib import Path
-from typing import Any
+from typing import Any, TypeVar
 
 from utils.exclusion_rules import RuleTokens
 
@@ -33,6 +33,8 @@ PROJECTS_CACHE_FILE = CACHE_DIR / "projects.json"
 COMPOSER_MAP_CACHE_FILE = CACHE_DIR / "composer-id-to-ws.json"
 INVALID_WORKSPACE_ALIASES_CACHE_FILE = CACHE_DIR / "invalid-workspace-aliases.json"
 TAB_SUMMARIES_PREFIX = "tab-summaries-"
+
+T = TypeVar("T")
 
 
 def nocache_enabled(*, request_nocache: bool = False) -> bool:
@@ -166,11 +168,6 @@ def _read_cache_file_unlocked(path: Path | str) -> dict[str, Any] | None:
         return None
 
 
-def _read_cache_file(path: Path | str) -> dict[str, Any] | None:
-    with _summary_cache_lock:
-        return _read_cache_file_unlocked(path)
-
-
 def _write_cache_file_unlocked(path: Path | str, payload: dict[str, Any]) -> None:
     p = Path(path)
     try:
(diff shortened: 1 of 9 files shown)


bradjin8 (approved):
## Optional follow-ups
### services/summary_cache.py
Global lock serializes fingerprint I/O across unrelated workspaces; per-cache-key locks if threaded latency regresses

### tests/test_summary_cache_concurrency.py
Only projects cache tested under concurrency; other three types share _get_or_build_cached so risk is low

services/summary_cache.py
```diff
@@ -200,18 +233,72 @@ def set_cached_projects(
...
+
+    built = build_fn()
+
+    with _summary_cache_lock:
+        fingerprint = _workspace_storage_fingerprint(workspace_path, workspace_entries, rules)
+        hit = get_unlocked(fingerprint)
+        if hit is not None:
+            return hit
+        if should_cache is None or should_cache(built):
+            set_unlocked(fingerprint, built)
```
wpak-ai: Persist the built value under the pre-build fingerprint, or skip the write entirely when the post-build fingerprint differs from the pre-build one, so a payload is never stored under a fingerprint describing storage bytes it never read

wpak-ai (changes requested):
1. `services/export_engine.py:87`, `services/export_engine.py:146`, `services/workspace_listing.py:82` - Delete the `fingerprint` field from `WorkspaceOrchestration` and stop computing it in `prepare_workspace_orchestration`, now that the cache key is derived entirely inside `_get_or_build_cached`

2. `services/summary_cache.py:118`, `services/export_engine.py:146`, `services/search_index.py:131` - Make the new resolver public by dropping the underscore from `_workspace_storage_fingerprint`, then call it from `prepare_workspace_orchestration` and `search_index._storage_fingerprint` instead of their own copies of the resolve-then-fingerprint block

services/summary_cache.py
clean6378-max-it: fixed.

wpak-ai (approved):

PR #159, opened 2026-07-30
feat: tag-triggered release workflow for wheel, sdist and Windows bundle

clean6378-max-it:
Closes #155 

Adds a release workflow that runs when you push a version tag. It builds the Python packages and the Windows zip, checks that all three files are there, then attaches them to the GitHub Release. You can also trigger the build jobs manually without publishing. CI now installs pywebview for the Windows PyInstaller step so it matches what the release job ships.

I dry-ran the build locally on 0.2.0: wheel about 154 KiB, sdist about 225 KiB, Windows zip about 36 MiB. The exe passed a --help smoke test. A test tag on a fork is still the way to confirm publish end-to-end before we cut 0.2.1.

## Summary by CodeRabbit

* **New Features**
  * Release workflow can be triggered manually for build verification; publishing is still limited to `v*` tag pushes.
  * Added a check to ensure the expected release artifacts (wheel, sdist, and Windows bundle) are present before publishing.

* **Bug Fixes**
  * Improved Windows desktop packaging by ensuring the required webview dependency is installed during the PyInstaller build.

* **Documentation**
  * Refined “GitHub Releases” instructions for matching `pyproject.toml` version with the pushed tag, and updated related wording.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,6 +1,6 @@
 name: Release
 
-# Fires when an annotated version tag is pushed (for example v0.2.1).
+# Fires when a version tag is pushed (for example v0.2.1).
 # Builds hatchling wheel + sdist and a Windows PyInstaller zip, then attaches
 # all three as GitHub Release assets via softprops/action-gh-release.
 #
@@ -18,7 +18,7 @@ on:
       - "v*"
 
 permissions:
-  contents: write
+  contents: read
 
 concurrency:
   group: release-${{ github.ref }}
@@ -101,6 +101,8 @@ jobs:
     name: Publish GitHub Release assets
     needs: [build-python, build-windows]
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
     steps:
       - name: Download Python distributables
         uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093  # v4
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,17 +1,14 @@
 name: Release
 
-# Fires when a version tag is pushed (for example v0.2.1).
-# Builds hatchling wheel + sdist and a Windows PyInstaller zip, then attaches
-# all three as GitHub Release assets via softprops/action-gh-release.
+# v* tag push: build wheel, sdist, and Windows zip; publish attaches all three to the Release.
+# workflow_dispatch runs the build jobs only (no publish).
 #
-# Expected asset names (version comes from pyproject.toml at the tagged commit):
+# Asset names (version from pyproject.toml at the tagged commit):
 #   cppa_cursor_browser-<version>-py3-none-any.whl   (~154 KiB at 0.2.0)
-#   cppa_cursor_browser-<version>.tar.gz               (~225 KiB at 0.2.0)
-#   CursorChatBrowser-windows.zip                      (PyInstaller onedir bundle)
+#   cppa_cursor_browser-<version>.tar.gz             (~225 KiB at 0.2.0)
+#   CursorChatBrowser-windows.zip
 #
-# Fork verification: push a test tag (for example v0.0.0-test) and confirm the
-# workflow attaches all three files to the resulting Release.
-# workflow_dispatch runs build-python + build-windows only (publish is tag-only).
+# Fork check: push v0.0.0-test and confirm all three files on the Release.
 
 on:
   push:
@@ -68,8 +65,7 @@ jobs:
           python-version: "3.12"
 
       - name: Install runtime dependencies
-        # Lock file for Flask/runtime deps; pywebview is the [desktop] extra and
-        # is required so the published exe can open its native window.
+        # requirements-lock.txt for runtime; pywebview for the desktop window.
         run: |
           python -m pip install --upgrade pip
           python -m pip install -r requirements-lock.txt
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,7 +8,7 @@ name: Release
 #   cppa_cursor_browser-<version>.tar.gz             (~225 KiB at 0.2.0)
 #   CursorChatBrowser-windows.zip
 #
-# Fork check: push v0.0.0-test and confirm all three files on the Release.
+# Fork rehearsal: push v0.2.0 (pyproject is already 0.2.0) and confirm all three assets on the Release.
 
 on:
   push:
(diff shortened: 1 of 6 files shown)


cursor-browser.spec
```diff
@@ -17,6 +17,7 @@ a = Analysis(
         (str(src / "static"), "static"),
     ],
     hiddenimports=[
+        "webview",
```
bradjin8: `webview` hiddenimport is correct for the published Windows bundle, but it only gets bundled when `pywebview` is installed at build time (release.yml does this; tests.yml does not yet). Worth a one-line comment here so future edits don't assume hiddenimport alone is sufficient.

.github/workflows/release.yml
```diff
@@ -0,0 +1,122 @@
...
+          name: python-distributables
+          path: release-assets
+
+      - name: Download Windows bundle
+        uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093  # v4
+        with:
+          name: windows-bundle
+          path: release-assets
+
+      - name: Attach artifacts to Release
```
bradjin8, line 152: Optional hardening before attach: add a step that asserts `release-assets/` contains exactly one `.whl`, one `.tar.gz`, and `CursorChatBrowser-windows.zip` (e.g. `ls -la release-assets/` + `test $(ls release-assets | wc -l) -eq 3`). Cheap guard against a partial or mis-merged artifact download.

.github/workflows/release.yml
```diff
@@ -0,0 +1,122 @@
...
+#   cppa_cursor_browser-<version>.tar.gz               (~225 KiB at 0.2.0)
+#   CursorChatBrowser-windows.zip                      (PyInstaller onedir bundle)
+#
+# Fork verification: push a test tag (for example v0.0.0-test) and confirm the
+# workflow attaches all three files to the resulting Release.
+
+on:
+  push:
+    tags:
+      - "v*"
```
bradjin8, line 16: Issue #155 Verification allows a dry run as an alternative to a fork tag. Consider adding `workflow_dispatch:` alongside the `v*` tag trigger so maintainers can exercise build-python + build-windows (and optionally publish to a draft) without pushing a version tag. Not blocking if fork evidence is in the PR body.

README.md
```diff
@@ -263,6 +263,18 @@ Adding new optional fields to JSON responses, adding new CLI flags with sensible
...
+
+Pushing a tag matching `v*` (for example `v0.2.1`) runs [`.github/workflows/release.yml`](.github/workflows/release.yml), which attaches three assets to the GitHub Release:
+
+| Asset | Contents |
+|---|---|
+| `cppa_cursor_browser-<version>-py3-none-any.whl` | Installable wheel (hatchling build) |
+| `cppa_cursor_browser-<version>.tar.gz` | Source distribution |
+| `CursorChatBrowser-windows.zip` | Windows PyInstaller onedir bundle (`CursorChatBrowser.exe` plus supporting files) |
+
+At `0.2.0`, local `python -m build` produced a ~154 KiB wheel and ~225 KiB sdist; the Windows zip size depends on the locked dependency tree at tag time. Copy release notes from the matching `[version]` section in `CHANGELOG.md` when publishing the GitHub Release.
```
bradjin8: Helpful section. For is-10, consider one sentence that `pyproject.toml` `[project].version` must match the tag (e.g. tag `v0.2.1` → version `0.2.1`) before push, since hatchling names the wheel/sdist from pyproject, not from the git tag.

bradjin8 (changes requested):
### [Should fix] Issue #155 acceptance criteria and Verification section ask for a fork tag push or dry run with the resulting artifact names and sizes recorded in the PR body. 

The workflow header (release.yml:7–10) and README table document expected names and local 0.2.0 wheel/sdist sizes, but this PR body only has local sanity-check commands - no fork Actions run link or actual uploaded asset sizes (especially CursorChatBrowser-windows.zip).

Please add a short "Release workflow verification" section here with either:
- a fork tag push (e.g. v0.0.0-test) + link to the Actions run and the three asset filenames/sizes from the Release, or
- a dry-run summary from a local `python -m build` + `pyinstaller` run including the Windows zip size at the locked dependency tree.

Without that, we won't know the publish job works end-to-end before is-10.

### [Nice to have] .github/workflows/tests.yml:123
`cursor-browser.spec` now lists `webview` as a hiddenimport (see spec line 20), and the release workflow correctly installs `pywebview` before PyInstaller. The unittest matrix Windows PyInstaller steps (123–129) still build without `pywebview`, so CI only smoke-tests `--help` and does not exercise the same bundle graph as the published Release asset.

Consider adding `python -m pip install 'pywebview>=5.0,<7'` to the Windows 3.12 PyInstaller block (matching release.yml:68–74) so CI validates the desktop bundle we actually ship.

cursor-browser.spec
clean6378-max-it: Added a one-liner: webview only lands in the bundle when pywebview is installed at build time.

.github/workflows/release.yml
clean6378-max-it: Now checks release-assets/ has exactly one .whl, one .tar.gz, and CursorChatBrowser-windows.zip before attach.

.github/workflows/release.yml
clean6378-max-it: Added. workflow_dispatch runs both build jobs

README.md
clean6378-max-it: Added a sentence that pyproject.toml [project].version must match the tag before push.

bradjin8 (approved):

wpak-ai (changes requested):
1. `.github/workflows/release.yml:80-81`, `.github/workflows/tests.yml:136-138`: replace both `--help` smoke tests with an assertion that the pywebview payload actually landed, e.g. fail when `dist\CursorChatBrowser\_internal\webview\lib` is absent, and add an explicit `shell: pwsh` to the release step (the step gating the Windows binary cannot fail for two independent reasons, `launcher.py:24` exits on `--help` before `import webview` at line 27 while PyInstaller logs an unresolved hidden import and exits 0, and `console=False` at `cursor-browser.spec:58` makes the exe GUI-subsystem so the default `pwsh` never waits or sets `$LASTEXITCODE`, so this PR's own fix has no test that can observe it and a windowless exe would publish silently)

2. `.github/workflows/release.yml:11`: replace the `v0.0.0-test` fork rehearsal with `v0.2.0`, which matches pyproject today and exercises the full path including publish (the documented recipe can never produce a Release, because the verify step derives the expected wheel name from the tag while hatchling names it from pyproject's `0.2.0`, and `Version("0.0.0-test")` raises `InvalidVersion` so retagging pyproject to match is impossible, this is the recipe for issue #155's Part E acceptance criterion)

3. `pyproject.toml`, `.github/workflows/release.yml:123`, `README.md:268`, `CHANGELOG.md:113`: add a preflight step at the top of `build-python` comparing `${GITHUB_REF_NAME#v}` against `[project].version` and failing with a message naming both values (the tag-equals-pyproject invariant is instructed in the README, stored in pyproject, mirrored by the CHANGELOG headings and compare links, and enforced only by a bare `test -f` in the final job, so editing one surface without the others costs both builds and surfaces as an unexplained non-zero exit)

4. `.github/workflows/release.yml:72`, `.github/workflows/tests.yml:126`, `pyproject.toml:29`: add a check that the pywebview range in both workflows matches `[project.optional-dependencies].desktop`, since installing the extra directly would bypass `requirements-lock.txt` and is not a clean substitute (`pywebview>=5.0,<7` is written out three times character-identically and the comment at `tests.yml:123` asserts the copies match, but nothing enforces it, so a bump in pyproject would silently leave the shipped Windows bundle built against the old range)

wpak-ai (approved):

